PR #5, opened 2013-11-19
c++ interpreter bridge using fibers

volgy:


lattmann:
We may use these changes in the future, but not in the short term.

PR #6, opened 2013-12-18
Fixes and additions to the Common and Tasync modules needed for C++ bridge

azeey:
These commits are included in the C++ bridge branch but I wanted to extract them so that they can be merged sooner. While the changes in the Tasync module are only needed in the C++ bridge, the changes in the Common module might also be useful for other parts webgme.

### DIFF
--- a/core/tasync.js
+++ b/core/tasync.js
@@ -725,6 +725,10 @@
 		return first;
 	}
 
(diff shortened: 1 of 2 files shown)


lattmann:
We may use these changes in the future, but not in the short term.

PR #108, opened 2014-12-10
fix #106: need to wait for server to get branch creation message

### LINKED ISSUE #106: Creating a branch then loading it immediately fails with "there is no such branch!"
Seems to be because `storage/failsafe.js`'s `setBranchHash` calls `callback` immediately.

ksmyth:
This PR looks like it is working


ksmyth:
I don't remember how to repro the issue this fixed. I think the recent fsync changes may have obviated these changes

Issue #78, opened 2014-07-24
Redirect to requested page after login


ksmyth:
I have authentication turned on for WebGME. If I go to e.g. `/rest/blob`, it redirects to `/login`. After I log in, it goes to the WebGME editor. I'd like for it to go to `/rest/blob` instead.

Suggestion: response for unauthenticated `/rest/blob` sends `Location: /login?redirect=/rest/blob`, and `/login` reads the redirect query parameter.

This is needed for the domain-tools executor worker.

PR #147, opened 2015-02-11
#78 login redirection have been implemented differently

kecso:
-> login.html have to be redesigned/refactored
login.html updated
now /login/gme is used instead of post request
redirect url is saved in a query parameter (more bulletproof)
unfortunately username and password is exposed in query string so they have to be coded immediately

### DIFF
--- a/src/client/login.html
+++ b/src/client/login.html
@@ -69,7 +69,7 @@
 
     <body>
         <div class="container">
-            <form class="form-signin" action="/login" method="post">
+            <form class="form-signin">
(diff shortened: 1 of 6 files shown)


kecso:
thanks for the comments!
I will change the code then.
1. unfortunately I was not aware of the existence of those functions (and that they are available under nodejs as well)
2. I totally forgot that the URL is never encoded - I thought for the time being it doesn't matter where we put those values as they are not encoded anyway, but that is not true under https -
thanks again.

PR #174, opened 2015-02-23
Rewrite authentication to use mongodb directly

ksmyth:


src/bin/usermanager.js
```diff
@@ -13,40 +14,38 @@ if (typeof define !== "function") {
...
+            })
+            .finally(gmeauth.unload);
 
-        function main () {
+        function main() {
             //var args = COMMON.getParameters(null);
             //console.log(args);
 
-            if (COMMON.getParameters("help") !== null) {
+            var usage = function () {
```
lattmann: We should use the commander module in the future.

Issue #168, opened 2015-02-23
Use bcrypt for authentication credentials


ksmyth:
Storing plaintext passwords is bad security practice. Use bcrypt hashing instead.

PR #176, opened 2015-02-23
Use bcrypt for authentication

ksmyth:
Storing plaintext passwords is bad security practice. Using MD5/SHA1/SHA256, even with salt, is bad security practice. Use bcrypt. http://codahale.com/how-to-safely-store-a-password/


ksmyth:
This is issue #168

PR #184, opened 2015-02-25
Issue/169 autorouter section has blocked edge assert failure

brollb:
Fixed autorouter bug caused by floating point addition error.


pmeijer:
Is this going into v0.7.x too?

lattmann:
It was branched from master, and master was merged into this branch. I would target 0.8.0 with all autorouter fixes.

Issue #193, opened 2015-03-02
Fix blob issues with node 0.11


pmeijer:
Tests are failing,
https://travis-ci.org/webgme/webgme/jobs/52490450

PR #194, opened 2015-03-02
Issue/192 coreplugins

pmeijer:
This branch includes issue #193


.gitignore
```diff
@@ -31,3 +31,7 @@ blob-local-storage
 *.swp
 
 webgme-*.tgz
+
+src/middleware/executor/worker/node_modules
```
lattmann: change /node_modules to node_modules, then we do not need this line.

pmeijer: Done

PR #212, opened 2015-03-04
On failed auth, redirect to /login. Also set username and redirect, and ...

ksmyth:
...display an error message

### DIFF
--- a/src/client/login.html
+++ b/src/client/login.html
@@ -46,6 +46,14 @@
                 padding: 7px 9px;
             }
 
+            .form-error {
+                border: 1px solid #dd4b39;
+            }
+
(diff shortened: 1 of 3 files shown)


src/client/login.html
```diff
@@ -73,11 +81,12 @@
                 <h2 class="form-signin-heading">WebGME sign in</h2>
                 <div class="row">
                     <div class="col-md-6">
-                        <input name="username" type="text" autocomplete="off" autofocus="autofocus" class="input-block-level" placeholder="Username">
```
lattmann: Why do we need to remove autofocus? If we visit the page i can just start typing...

Issue #228, opened 2015-03-12
Url query for node to load should be node and it should use the config if defined there.


pmeijer:
Example;

http://localhost:9091/?project=TT&branch=b1&node=/1

will load project "TT", branch "b1" and node with id "/1".

http://localhost:9091/?project=TT&node=/1

will load project "TT", branch "master" and node with id "/1".

In the config it is here;

    
client: {
        appDir: path.join(__dirname, '../src/client'),
        // Used in client/WebGME.js to load initial project.
        defaultProject: {
            name: null,
            branch: null,
            node: null
        }
    },

PR #229, opened 2015-03-12
#228 update query and use configuration.

pmeijer:


lattmann:
Looks good, we have to add it to the 0.8.0 release notes 'breaking changes' section.

PR #253, opened 2015-03-18
Closes #225 upgrade superagent 0.18.2 to 0.21.0

### LINKED ISSUE #225: Upgrade superagent
We have a custom superagent.js for the client to support binary downloads. A new release of superagent supports this:
https://github.com/visionmedia/superagent/pull/566

BlobClient.prototype.getObject may need to be changed for this API

(I would get the webgme-domain-tools karma tests that test the blob working before doing anything)

lattmann:


lattmann:
I am sorry, i did not realize we had special superagent version. All tests passed, so I am wondering if we have any action item here.

What would you prefer to do? We can reopen the #225 issue, if the current code is not functional.

ksmyth:
The tests work, and I think they cover the use cases. So I am satisfied that it works correctly

PR #285, opened 2015-03-31
Fixes #119 blob client keep alive under node and node webkit.

### LINKED ISSUE #119: Use keepalive in BlobClient
Uploading many small files can be slow with the TCP creation/teardown overhead for each file, especially if the GME server is far away from the client. (Double especially if using https)

This stackoverflow answer has some hints on how to do it with superagent:
http://stackoverflow.com/questions/20743546/keep-alive-connection-with-superagent

lattmann:


lattmann:
Should be ready now...

lattmann:
https://github.com/node-modules/agentkeepalive/pull/18

ksmyth:
Looks good. Thanks for your hard work on this.

I don't know if `maxSockets: 100` may be too high, but due to `blob/Artifact.js` throttling requests, it probably doesn't matter.

PR #313, opened 2015-04-08
The parsing of a huge json file (from blob) will not cause stack overflow

kecso:


src/common/blob/BlobClient.js
```diff
@@ -258,7 +258,12 @@ define(['blob/Artifact', 'blob/BlobMetadata', 'superagent'], function (Artifact,
...
                     if (contentType === 'application/json') {
                         var utf8ArrayToString = function (uintArray) {
-                            return decodeURIComponent(escape(String.fromCharCode.apply(null, uintArray)));
+                            var inputString = '',
+                                i;
+                            for (i = 0; i < uintArray.byteLength; i++) {
+                                inputString += String.fromCharCode(uintArray[i]);
+                            }
+                            return decodeURIComponent(escape(inputString));
                         };
```
pmeijer: Couldn't this be very inefficient to do for each byte?
Maybe convert in chunks instead..

ksmyth: Funny you should ask, I made this:
http://jsperf.com/arr-fromcharcode

By changing the chunk size, I was able to make all of the tests run with different timing. I could not make it produce reliable results

pmeijer: Cool,
I tried with bigger sizes and couldn't get reliable results either. 
The one-by-one "seemed" to perform best on average though.

PR #317, opened 2015-04-09
Removed ArPoint rounding. Fixes #306

### LINKED ISSUE #306: Custom Path Point Ordering
When I add a custom path point, it connects them based on the order of creation not ordered based on placement.

brollb:
Also had to update start/end point creation in the port


brollb:
And added test, of course

PR #320, opened 2015-04-10
Closes #316 storage and core logging

### LINKED ISSUE #316: Add/revise logging in storage and core layer.

pmeijer:
The branch also introduced a common regexp file.

### DIFF
--- a/src/bin/import.js
+++ b/src/bin/import.js
@@ -8,23 +8,22 @@
 
 var webgme = require('../../webgme'),
     program = require('commander'),
-    BRANCH_REGEXP = new RegExp('^[0-9a-zA-Z_]*$'),
     FS = require('fs'),
     path = require('path'),
     openContext,
     Serialization,
-    jsonProject;
+    jsonProject,
+    gmeConfig = require(path.join(process.cwd(), 'config')),
+    logger = webgme.Logger.create('gme:bin:import', gmeConfig.bin.log),
+    REGEXP = webgme.REGEXP,
+    openContext = webgme.openContext,
+    Serialization = webgme.serializer;
 
 
-openContext = webgme.openContext;
-Serialization = webgme.serializer;
-
-
-var importProject = function (Storage, gmeConfig, projectId, jsonProject, branchName, overwrite, callback) {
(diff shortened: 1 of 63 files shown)


src/bin/import.js
```diff
@@ -40,7 +39,7 @@ var importProject = function (Storage, gmeConfig, projectId, jsonProject, branch
         };
 
 
-    storage = new Storage({globConf: gmeConfig, log: logger.fork('storage')});
+    storage = new Storage({globConf: _gmeConfig, logger: logger.fork('storage')});
```
kecso: why the gmeConfig changed to _gmeConfig? (I mean the definition of it doesn't changed)
the same goes for line 52.

src/common/core/core.js
```diff
@@ -16,20 +19,33 @@ define([
...
+        coreLayers.push(CoreDiff);
+        coreLayers.push(TreeLoader);
 
-        var coreCon = new TreeLoader(new CoreDiff(new MetaCore(new Constraint(new Guid(new Set(new NullPtr(new Type(new NullPtr(new CoreRel(new CoreTree(storage, options)))))))))));
+        coreCon = coreLayers.reduce(function (inner, Class) {
+            return new Class(inner, options);
+        }, new CoreTree(storage, options));
 
-        if(options.usertype === 'tasync'){
+        if (options.usertype === 'tasync') {
```
kecso: I would put this option check before the reducing, and add Unwrap to the array if needed.

src/bin/import.js
```diff
@@ -8,23 +8,22 @@
...
+    openContext = webgme.openContext,
+    Serialization = webgme.serializer;
 
 
-openContext = webgme.openContext;
-Serialization = webgme.serializer;
-
-
-var importProject = function (Storage, gmeConfig, projectId, jsonProject, branchName, overwrite, callback) {
+var importProject = function (Storage, _gmeConfig, projectId, jsonProject, branchName, overwrite, callback) {
```
pmeijer: To distinguish between the global gmeConfig (used when the script is running as bin) and the configuration passed via importProject (e.g. in tests).

PR #324, opened 2015-04-13
Fix bug where system.indexes shows up in project list

ksmyth:


src/common/regexp.js
```diff
@@ -9,7 +9,8 @@ define([], function () {
     var HASH = new RegExp('^#[0-9a-zA-Z_]*$'),
         BRANCH = new RegExp('^[0-9a-zA-Z_]*$'),
         RAW_BRANCH = new RegExp('^\\*[0-9a-zA-Z_]*$'),// This is how it's stored in mongodb, i.e. with a prefixed *.
-        PROJECT = new RegExp('^[0-9a-zA-Z_]*$'),
+        PROJECT = new RegExp('^(?!system\\.)(?!_)[0-9a-zA-Z_]*$'), // project name may not start with system. or _
```
ksmyth: Maybe [0-9a-zA-Z] should have + instead of * ?

lattmann: Yes, we will change how we store the projects and the regexp for the next release 0.10.0 anyways.

PR #329, opened 2015-04-15
Closes #326 Don't rely on order of activeObject and activeSelection.

### LINKED ISSUE #326: Property editor is not updated
When an empty spot is clicked in the diagram designer, the property editor shows the parent of the current object instead of the current object. If the current object has children, selecting a child sometimes fixes the problem.

issue

pmeijer:
Instead check if activeSelection is within or equals activeObject.


pmeijer:
This should go into master too.

PR #356, opened 2015-05-01
Add Dockerfile to run webgme in a docker container.

lattmann:


Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+ENV PATH      $NVM_DIR/v$NODE_VERSION/bin:$PATH
+
+
+
+RUN echo smallfiles = true >> /etc/mongodb.conf
+
+RUN mkdir -p /project
+ADD package.json /project/package.json
+RUN cd /project
+
```
ksmyth: Running `npm install` here reduces the container startup time, and reduces container build time when package.json hasn't changes (which should be 90%+ of commits)

Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+ENV NODE_PATH $NVM_DIR/v$NODE_VERSION/lib/node_modules
+ENV PATH      $NVM_DIR/v$NODE_VERSION/bin:$PATH
+
+
+
+RUN echo smallfiles = true >> /etc/mongodb.conf
+
+RUN mkdir -p /project
+ADD package.json /project/package.json
+RUN cd /project
```
ksmyth: How is this different than the `WORKDIR` directive?

lattmann:
Changed according to your suggestions. Tested and it worked. Thanks. Should be ready to merge.

PR #358, opened 2015-05-05
Fix 'uncaughtException: Can't set headers after they are sent.' error

ksmyth:


lattmann:
I have to fix all (6) failing tests for this before i merge it.

ksmyth:
Sorry about the failing tests. You could check e.errno === ENOENT and send 404 in that case

PR #359, opened 2015-05-05
storage/mongo: dont ASSERT on user-provided input

ksmyth:
I am seeing server crashes from invalid input


kecso:
I think we should also replace the ASSERT(typeof callback = 'function'), as that is also 'user input'

lattmann:
@kecso would you please fix the failing test?

ksmyth:
By "user", I meant "someone accessing the server via socket.io", not a user of storage/mongo.js

PR #364, opened 2015-05-14
Added destroy() on CRM on panel destroy. Fixes #363

### LINKED ISSUE #363: ConnectionRouteManager not being destroyed on panel change
When switching from the Composition to the MetaAspect, the connection route manager is not being destroyed (though the parent diagram designer is).

brollb:


src/client/js/Widgets/DiagramDesigner/DiagramDesignerWidget.js
```diff
@@ -359,6 +359,7 @@ define([
     DiagramDesignerWidget.prototype.destroy = function () {
         this.__loader.destroy();
         this._removeToolbarItems();
+        this.connectionRouteManager.destroy();
         //TODO: what about item and connection destroys????
```
pmeijer: Is this TODO related?

brollb: According to git blame, Robert added that line May 14, 2013... Probably should leave it for now

PR #367, opened 2015-05-15
BlobClient: use globalAgent's certificate authorities (unless the BloClient user passed a custom list)

ksmyth:
This fixes node executor_workers with a custom ca (broken since keepaliveagent was introduced)


lattmann:
Blob tests are continuously failing, I have restarted them, but still... I have to look at the tests and fix them before i merge this issue.

ksmyth:
I pushed an update

lattmann:
Thanks.

PR #389, opened 2015-06-11
Closes #386 automerge

### LINKED ISSUE #386: automerge option to the ui
we should put the branch automerge possibility to the ui:
- if there is no conflict the target branch could be updated
- if there is some conflict, then we could show the user what needs to be dealt with
- a worker could do the job

pmeijer:
This should go both to v0.11.x and master


lattmann:
Merged to `v0.11.x` and `master`.

PR #390, opened 2015-06-11
Better malformed hash error message

ksmyth:


lattmann:
Should this go into `v0.11.1` or to the master only? PR was created against the master.

ksmyth:
Master only is fine

PR #396, opened 2015-06-12
Client plugins

pmeijer:


pmeijer:
This adds config.plugin.displayAll (false by default).

PR #426, opened 2015-06-26
Support BlobClient.getSubObject in tests

ksmyth:


lattmann:
I do not like that the callback is not the last parameter, but it is ok for now.

PR #438, opened 2015-07-01
Closes #434 specify branchName on selectProject in client.

### LINKED ISSUE #434: Client improvement: be able to open a project with a specific branch
Right now, the client always loads 'master' (or if it doesn't exist, one at random). It would be nice to be able to choose, so we don't pay the performance penalty

pmeijer:


src/client/js/client.js
```diff
@@ -333,12 +333,14 @@ define([
...
         };
 
-        this.selectProject = function (projectId, callback) {
+        this.selectProject = function (projectId, branchName, callback) {
             if (isConnected() === false) {
                 callback(new Error('There is no open database connection!'));
             }
             var prevProjectId,
-                branchToOpen = 'master';
+                branchToOpen = branchName || 'master';
```
ksmyth: Can keep API compat: if (callback === undefined && typeof branchName === 'Function') { callback = branchName; branchName = undefined }

PR #473, opened 2015-08-06
Plugin and userproject

pmeijer:
### Storage ###
- `common/storage/project.js`: supports promises.
- `server/safestorage.js`: `openProject` and `createProject` returns `UserProject`. (`_getProject` exposes lower level project.)
- `server/storage.js`: `setBranchHash` also returns `hash` (as `makeCommit`).
- `deleteBranch` exposed on project API.
- `editorstorage.js`: deletes projects and branches pointers AFTER closed on server.

### run_plugin ###
- uses mandatory args for pluginName and projectName.
- better user feedback.

### InterpreterManager  ###
- makes sure client is in sync when invoking plugin.
- returns with pluginResult on errors (where it didn't before). Exceptions thrown on client side still not handled.

Added coverage and minor fixes to coreplugins.

AddOn bug fix when starting multiple times.


src/common/storage/project/interface.js
```diff
@@ -39,6 +40,10 @@ define([
             throw new Error('createBranch must be overridden in derived class');
         };
 
+        this.deleteBranch = function (branchName, oldHash, callback) {
+            throw new Error('createBranch must be overridden in derived class');
```
ksmyth: Fix copy/paste error

test/bin/run_plugin.spec.js
```diff
@@ -43,19 +40,16 @@ describe('Run plugin CLI', function () {
...
-                    storage.closeDatabase(),
-                    gmeAuth.unload()
-                ]);
-            })
-            .nodeify(done);
+        Q.allSettled([
+            storage.closeDatabase(),
+            gmeAuth.unload()
+        ])
+        .nodeify(done);
```
ksmyth: this will lose any error raised by closeDatabase or unload

e.g.
    > q.allSettled([q.reject('asdf')]).nodeify(function (err, res) { console.log(err + ': ' + res) })
    undefined
    > null: [object Object]

pmeijer: True that, I will change it to q.all.

PR #508, opened 2015-08-27
Closes #482 Add reference counting in mongo.

### LINKED ISSUE #482: mongo.js openDatabase should be refcounted
mongo.js:

    if (mongo === null) {
        logger.debug('Connecting to mongo...');
        // connect to mongo
        mongodb.MongoClient.connect(gmeConfig.mongo.uri, gmeConfig.mongo.options, function (err, db) {
            if (!err && db) {
                mongo = db;

1. What happens if you call openDatabase twice in a row? (two connections are opened, one is lost and never closed)
2. What happens after this:

    client1Storage.openDatabase()
    // elsewhere
    client2Storage.openDatabase()
    client1Storage.closeDatabase()
    client2Storage.loadObjects(...)
    client2Storage.closeDatabase()

(client2Storage throws an error that the database hasn't been opened, even though it has and it hasn't yet called closeDatabase())

Suggested solution: keep a refcount in mongo.js. openDatabase increments it; if the ++refcount is 1, create a promise. If ++refcount > 1, return that promise. closeDatabase decrements it; if --refcount is 0, the mongo connection is closed.

pmeijer:
See my comment at #482 .

### DIFF
--- a/src/server/storage/mongo.js
+++ b/src/server/storage/mongo.js
@@ -18,6 +18,9 @@ var mongodb = require('mongodb'),
 
 function Mongo(mainLogger, gmeConfig) {
     var mongo = null,
+        connectionCnt = 0,
+        connectDeferred,
+        disconnectDeferred,
         logger = mainLogger.fork('mongo');
 
     /**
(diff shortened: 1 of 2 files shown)


src/server/storage/mongo.js
```diff
@@ -330,47 +333,66 @@ function Mongo(mainLogger, gmeConfig) {
...
+                        connectDeferred.resolve();
+                    } else {
+                        mongo = null;
+                        connectionCnt -= 1;
+                        logger.error('Failed to connect.', {metadata: err});
+                        connectDeferred.reject(err);
+                    }
+                });
+            } else {
+                logger.debug('Count is 1 but mongo is not null');
```
lattmann: Who is resolving the connectDeferred here? Will the callback be called if the openDatabase is called twice?

src/server/storage/mongo.js
```diff
@@ -330,47 +333,66 @@ function Mongo(mainLogger, gmeConfig) {
...
+            }
         } else {
             logger.debug('Reusing mongo connection.');
```
lattmann: Who is resolving the connectDeferred here? Will the callback be called if the openDatabase is called twice?

PR #510, opened 2015-08-27
Closes #504 create project API

### LINKED ISSUE #504: Allow project creation on the REST API
There should be a web API call to create project from seed, identical to the function provided by the client interface.

kecso:


src/server/api/index.js
```diff
@@ -573,6 +573,29 @@ function createAPI(app, mountPath, middlewareOpts) {
     });
 
 
+    router.put('/projects/:ownerId/:projectName', function (req, res, next) {
+        var userId = getUserId(req),
+            command = req.body;
```
lattmann: Add a comment here what kind of fields should be/can be passed by the body.

Like type is mandatory and seedName?
Add two examples in the comment.

lattmann:
There are no tests for `type: 'db'`. We should have at least one test.

lattmann:
Thanks for the changes, they look good.

PR #523, opened 2015-08-31
Fixes #515

### LINKED ISSUE #515: Exporting project / library under Safari fails

kecso:
Changing parameters of download links (target='_self') so it work until Safari as well

### DIFF
--- a/src/client/js/Utils/SaveToDisk.js
+++ b/src/client/js/Utils/SaveToDisk.js
@@ -1,0 +1,34 @@
+/*globals define*/
+/*jshint node: true, browser: true, bitwise: false*/
+
+/**
+ * @author kecso / https://github.com/kecso
+ */
+define(function () {
+    'use strict';
+
+    var saveToDisk = function (fileURL, fileName) {
+        // for non-IE
+        if (!window.ActiveXObject) {
+            var save = document.createElement('a');
+            save.href = fileURL;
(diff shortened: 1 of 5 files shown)


src/client/js/Utils/SaveToDisk.js
```diff
@@ -0,0 +1,34 @@
+/*globals define*/
+/*jshint node: true, browser: true, bitwise: false*/
+
+/**
+ * @author kecso / https://github.com/kecso
+ */
+define(function () {
+    'use strict';
+
+    var saveToDisk = function (fileURL, fileName) {
```
pmeijer: Was this code snippet coming from some external source, e.g. stack overflow? If so it's good to add a link.

src/server/standalone.js
```diff
@@ -852,24 +852,20 @@ function StandAloneServer(gmeConfig) {
                 logger.error('worker/simpleResult err', err);
                 res.sendStatus(500);
             } else {
-                res.header('Content-Disposition', 'attachment;filename=\'' + filename + '\'');
+                res.header('Content-Disposition', 'attachment;filename=' + filename);
```
ksmyth: This will be wrong if filename contains `%` or non-ASCII. Use https://www.npmjs.com/package/content-disposition

lattmann: +1

PR #528, opened 2015-09-02
Closes #525 cache non persisted

### LINKED ISSUE #525: Clearing out cache before all core-objects have been persisted can lead to hash not found.
Proposed solution: Store all non-persisted core-objects in a separate map that is cleared when the server has inserted them in the database.

pmeijer:


src/common/storage/storageclasses/editorstorage.js
```diff
@@ -356,6 +359,20 @@ define([
 
             if (project) {
                 project.insertObject(commitData.commitObject);
+                commitId = GUID();
```
kecso: Could we reuse the actual id of the commit?

pmeijer: Sure, good idea. Will fix it.

pmeijer:
Made the change, please merge

PR #530, opened 2015-09-03
Fixes socket.io transport issue

kecso:
Rollback to socket.io version 1.3.2 as later versions not able to setup websocket transport for communication under Safari and polling has problems with 'attachment' http requests during polling session.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,8 +54,8 @@
     "process": "=0.11.1",
     "q": "=1.4.1",
     "requirejs": "=2.1.18",
-    "socket.io": "lattmann/socket.io#3a824239cbe100225f1721e8a053a85fbd289dce",
-    "socket.io-client": "lattmann/socket.io-client#bb0baf236414e05a3548c1003b3a5f13a8042188",
+    "socket.io": "=1.3.2",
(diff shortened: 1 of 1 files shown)


lattmann:
Unfortunately, pointing to socket.io 1.3.2 will not include my bug fix for socket io. I can merge this since the tests are succeeding, but I have a feeling that something is still broken.

https://github.com/lattmann/socket.io-client/commit/bb0baf236414e05a3548c1003b3a5f13a8042188

kecso:
Sorry, I thought that the 'onack' issue was introduced in 1.3.3 .
Please let me know how we can confirm that socket.io has or not the 'onack' issue.

lattmann:
Looking at the source code?

kecso:
it does not have the correction, but that is not the confirmation of whether it has the same fault.
How should we proceed?
1, make the 'onack' correction into 1.3.2
2, try to test if the original issue arise in 1.3.2
3, still try to solve our main issue in 1.3.6 (the 'websocket' does not work and they not seem to fix it, the 'polling' also have its issues for long...)

lattmann:
Ok, you said you have tested it. I am going to merge it.

PR #531, opened 2015-09-03
Fixes #517 project dialog listens to delete/create events

### LINKED ISSUE #517: Project dialog doesn't react to project creation/deletion

kecso:
project dialog now reacts to project (delete/create) events


src/client/js/Dialogs/Projects/templates/ProjectsDialog.html
```diff
@@ -14,8 +14,6 @@
             </div>
             <div class="modal-footer">
                 <div class="panel-buttons">
-                    <button class="btn btn-default pull-left btn-refresh" title="Reload project list..."
```
pmeijer: Do we really want to remove this button? Don't see why we can't keep it.

PR #551, opened 2015-09-11
Added better box checking and warnings. Fixes #548

### LINKED ISSUE #548: ConnectionRouteManager3 Silent Errors
The ConnectionRouteManager3 sometimes receives events in the wrong order (resizing boxes before they exist) and handles this by checking that the box has been created by the Autorouter (specifically, that the `addBox` response has been received - if not, it sends an `addBox` command to the AR) and silently tolerates these out-of-order errors.

It should check that the `addBox` request has been sent (rather than received as this can lead to duplicate box creation) and should log errors about any out of order problems (resizing boxes that don't exist, etc).

brollb:


lattmann:
Fixes #548

PR #552, opened 2015-09-11
Fixes #393 event racing

### LINKED ISSUE #393: Update territories while loading in changes can cause inconsistencies in client.

kecso:
Now the pulling (the actual switch between the next and the current state of the project) will wait for the updateTerritory requests to finish. Parallely it loads everything, so the actual switch can happen in a single synchronous step and we should not loose any event.

### DIFF
--- a/src/client/js/client.js
+++ b/src/client/js/client.js
@@ -63,11 +63,24 @@ define([
                 inTransaction: false,
                 msg: '',
                 gHash: 0,
-                loadError: null
+                loadError: null,
+                ongoingTerritoryUpdateCounter: 0,
+                ongoingLoadPatternsCounter: 0,
+                pendingTerritoryUpdatePatterns: {},
+                loadingStatus: null,
+                inLoading: false,
+                loading: {
+                    rootHash: null,
+                    commitHash: null,
+                    next: null
+                }
+
             },
             monkeyPatchKey,
             nodeSetterFunctions,
-            addOnFunctions = new AddOn(state, storage, logger, gmeConfig);
+            addOnFunctions = new AddOn(state, storage, logger, gmeConfig),
+            loadPatternThrottled = TASYNC.throttle(loadPattern, 1); //magic number could be fine-tuned
+        //loadPatternThrottled = loadPattern; //magic number could be fine-tuned
 
         EventDispatcher.call(this);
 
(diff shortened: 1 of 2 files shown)


src/client/js/client.js
```diff
@@ -1409,67 +1303,211 @@ define([
         }
 
         function _updateTerritoryAllDone(guid, patterns, error) {
+            refreshMetaNodes(state.nodes, state.nodes);
+
             if (state.users[guid]) {
-                state.users[guid].PATTERNS = JSON.parse(JSON.stringify(patterns));
+                state.users[guid].PATTERNS = COPY(patterns);
```
pmeijer: Where is COPY defined?

src/client/js/client.js
```diff
@@ -1409,67 +1303,211 @@ define([
...
+
+            if (state.ongoingLoadPatternsCounter !== 0) {
+                logger.error('at the start of loading the counter should be zero!!! [' + state.ongoingTerritoryUpdateCounter + ']');
+            }
+
+            state.loadingStatus = null;
+            state.loadNodes = {};
+            state.loading.rootHash = newRootHash;
+            state.loading.commitHash = newCommitHash;
+            state.loading.next = callback || function () {
```
pmeijer: Doesn't loading require a callback?

src/client/js/client.js
```diff
@@ -1409,67 +1303,211 @@ define([
...
+        }
+
+        function loading(newRootHash, newCommitHash, callback) {
+            var i, j,
+                userIds,
+                patternPaths,
+                patternsToLoad = [];
+
+            if (state.ongoingLoadPatternsCounter !== 0) {
+                logger.error('at the start of loading the counter should be zero!!! [' + state.ongoingTerritoryUpdateCounter + ']');
```
pmeijer: If we're f-ed here, why not throw an Exception with the same error message. Some stack-trace couldn't heart..

src/client/js/client.js
```diff
@@ -1409,67 +1303,211 @@ define([
...
+                state.metaNodes[newPaths[i]] = newSource[newPaths[i]].node;
+            }
+
+            for (i = 0; i < pathsToRemove.length; i += 1) {
+                delete state.metaNodes[pathsToRemove[i]];
+            }
+        }
+
+        function switchStates() {
+            //it is safe now to move the loadNodes into nodes,
```
pmeijer: I would add more logging in general, but at least a debug message that we're switching state here.

src/client/js/client.js
```diff
@@ -63,11 +63,24 @@ define([
...
+                ongoingLoadPatternsCounter: 0,
+                pendingTerritoryUpdatePatterns: {},
+                loadingStatus: null,
+                inLoading: false,
+                loading: {
+                    rootHash: null,
+                    commitHash: null,
+                    next: null
+                }
+
```
pmeijer: Could these be grouped (together with users, nodes and loadNodes etc..)?

pmeijer:
Tamas were you sure that the new state variables shouldn't be reintialized when switching project or branch?
Can you also try to run it with debug = true? Afraid the newly introduced keys might not be stringified..

kecso:
Patrik, I've checked the situations regarding branch and project changes. In both scenarios, it is safer to wait for the updateTerritory requests to be resolved before switching to a new state (whether it is a new branch or project). This means, that if we would clear out any of the counters, than we could actually cause errors later (as there is a chance that some UT requests are ongoing at the time of switch).

PR #556, opened 2015-09-14
Refactor standalone.js

kecso:
As a first step, the Express.static has been taken into use for all static file resources, where the container directory is fixed as well.
Plugins, AddOns and External Services will come as a second step to completely eliminate the expressFileSending function.

### DIFF
--- a/src/server/standalone.js
+++ b/src/server/standalone.js
@@ -642,9 +642,7 @@ function StandAloneServer(gmeConfig) {
     // Basic authentication
 
     logger.debug('creating login routing rules for the static server');
-    __app.get('/', ensureAuthenticated, function (req, res) {
-        expressFileSending(res, __clientBaseDir + '/index.html');
-    });
+    __app.get('/', ensureAuthenticated, Express.static(__clientBaseDir));
     __app.get('/logout', function (req, res) {
         res.clearCookie('webgme');
         req.logout();
@@ -652,10 +650,9 @@ function StandAloneServer(gmeConfig) {
         delete req.session.udmId;
         res.redirect(__logoutUrl);
     });
-    __app.get('/login', function (req, res) {
-        res.location('/login');
-        expressFileSending(res, __clientBaseDir + '/login.html');
-    });
+
+    __app.get('/login', Express.static(__clientBaseDir, {extensions: ['html'], index: false}));
+
     __app.post('/login', function (req, res, next) {
         var queryParams = [],
             url = URL.parse(req.url, true);
@@ -752,6 +749,7 @@ function StandAloneServer(gmeConfig) {
         });
     });
 
+
     logger.debug('creating external library specific routing rules');
     __app.get(/^\/extlib\/.*/, ensureAuthenticated, function (req, res) {
         //first we try to give back the common extlib/modules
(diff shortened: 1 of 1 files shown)


src/server/standalone.js
```diff
@@ -642,9 +642,7 @@ function StandAloneServer(gmeConfig) {
     // Basic authentication
 
     logger.debug('creating login routing rules for the static server');
-    __app.get('/', ensureAuthenticated, function (req, res) {
-        expressFileSending(res, __clientBaseDir + '/index.html');
-    });
+    __app.get('/', ensureAuthenticated, Express.static(__clientBaseDir));
```
lattmann: I do not think this is right... does the UI load for you? Visiting only http://localhost:8888/ ? Or express sends the index.html by default?

src/server/standalone.js
```diff
@@ -774,33 +772,19 @@ function StandAloneServer(gmeConfig) {
...
-    __app.get(/^\/.*\.(_js|html|gif|png|bmp|svg|json|map)$/, ensureAuthenticated, function (req, res) {
-        //package.json
-        if (req.path === '/package.json') {
-            expressFileSending(res, Path.join(__baseDir, '..', req.path));
-        } else {
-            expressFileSending(res, Path.join(__clientBaseDir, req.path));
-        }
-    });
+    __app.get('/package.json', ensureAuthenticated, Express.static(Path.join(__baseDir, '..')));
+    __app.get(/^\/.*\.(_js|html|gif|png|bmp|svg|json|map)$/, ensureAuthenticated, Express.static(__clientBaseDir));
```
lattmann: Why do not we specify the extensions for express instead?

src/server/standalone.js
```diff
@@ -774,33 +772,19 @@ function StandAloneServer(gmeConfig) {
...
+    __app.get('/docs/*', Express.static(Path.join(__baseDir, '..'), {index: false}));
 
     __app.use('/rest/blob', BlobServer.createExpressBlob(middlewareOpts));
 
     //client contents - js/html/css
-    //stuff that considered not protected
-    __app.get(/^\/.*\.(css|ico|ttf|woff|js|cur)$/, function (req, res) {
-        expressFileSending(res, Path.join(__clientBaseDir, req.path));
-    });
+    __app.get(/^\/.*\.(css|ico|ttf|woff|js|cur)$/, Express.static(__clientBaseDir));
```
lattmann: Why do not we specify the extensions for express instead?

src/server/standalone.js
kecso: if you mean the static's extension parameter, then I think that would change the rule as it would try to load resources without extension as well.
Of course that probably not a big and bad change so if you feel, I can do it.

src/server/standalone.js
kecso: Yepp, the default behavior of static is that it enables directory lookup, and the default file is index.html

src/server/standalone.js
lattmann: Is that true for the login rule too? I just wanted to make sure it is consistent how we define the rules.

src/server/standalone.js
lattmann: Good thanks.

PR #557, opened 2015-09-14
Fixes #371 property browser now reacts to every selection change also in splitview

### LINKED ISSUE #371: Property editor is not updated in split view
In split view, after making some changes in one of the views, the property editor is not updated from the other view any more.

kecso:
the split view error was introduced with the #326, now we made a clear separation of active selection and active node
the treeBrowser could still follow the active node as in split view it is hard to track it, and can still cause confusion

### DIFF
--- a/src/client/js/Panels/PropertyEditor/PropertyEditorPanelController.js
+++ b/src/client/js/Panels/PropertyEditor/PropertyEditorPanelController.js
@@ -66,19 +66,10 @@ define(['js/logger',
             self._logger.debug('activeSelection changed', activeSelection);
             var activeNodeId = WebGMEGlobal.State.getActiveObject();
             if (activeSelection && activeSelection.length > 0) {
-                if (activeSelection.length === 1) {
-                    //https://github.com/webgme/webgme/issues/326
-                    if (self.startsWith(activeSelection[0], activeNodeId)) {
-                        self._selectedObjectsChanged(activeSelection);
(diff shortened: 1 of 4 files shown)


src/client/js/Panels/PropertyEditor/PropertyEditorPanelController.js
```diff
@@ -66,19 +66,10 @@ define(['js/logger',
             self._logger.debug('activeSelection changed', activeSelection);
             var activeNodeId = WebGMEGlobal.State.getActiveObject();
             if (activeSelection && activeSelection.length > 0) {
-                if (activeSelection.length === 1) {
-                    //https://github.com/webgme/webgme/issues/326
-                    if (self.startsWith(activeSelection[0], activeNodeId)) {
```
pmeijer: Is this issue not valid anymore? Oh I see you removed the pushs.. Did you confirm that it worked as expected though?

kecso: yes. I've checked the different scenarios and it seems to be consistent among them (that is why I removed the 'if nothing selected, let's select the activeNode part' from other UI parts). The original issue was coming from the confusion between the active selection and active node.
the only thing is that now if there is no selected item, the activeNode will be shown in the property editor, which can be a bit ambiguous in a split view, but otherwise clear in my opinion.

PR #567, opened 2015-09-14
Closes #543 constraints and meta-rules

### LINKED ISSUE #543: Refactor constraint checking and enable meta rule checking separately.
- [x] Move the meta rules to a separate function.
- [x] Make a constraint checker class for evaluation constraints (custom and the meta-rules).
- [x] Add separate entry point on GUI and worker for evaluating Meta-rules.
- [x] Check custom constraints w/o using AddOn.

pmeijer:
Tamas please check the common/users/metarules.js particularly.

### DIFF
--- a/config/config.default.js
+++ b/config/config.default.js
@@ -42,10 +42,6 @@ var path = require('path'),
             s3: {}
         },
 
-        core: {
-            enableCustomConstraints: false,
-        },
-
         client: {
             appDir: path.join(__dirname, '../src/client'),
             log: {
@@ -65,6 +61,10 @@ var path = require('path'),
                 node: null
             },
             defaultConnectionRouter: 'basic3' //'basic', 'basic2', 'basic3'
+        },
+
+        core: {
+            enableCustomConstraints: false,
         },
 
         debug: false,
--- a/config/config.default.js
+++ b/config/config.default.js
@@ -42,6 +42,10 @@ var path = require('path'),
             s3: {}
         },
 
+        core: {
+            enableCustomConstraints: false,
+        },
+
         client: {
             appDir: path.join(__dirname, '../src/client'),
             log: {
--- a/src/common/core/users/metarules.js
+++ b/src/common/core/users/metarules.js
@@ -8,6 +8,13 @@
 define(['q'], function (Q) {
     'use strict';
 
+    /**
+     *
+     * @param core
+     * @param node
+     * @param [callback]
+     * @returns {*}
+     */
     function checkMetaRules(core, node, callback) {
         var error = null,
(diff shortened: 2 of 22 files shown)


config/config.default.js
```diff
@@ -63,6 +63,10 @@ var path = require('path'),
             defaultConnectionRouter: 'basic3' //'basic', 'basic2', 'basic3'
         },
 
+        core: {
```
kecso: why is this called core? shouldn't we use 'cosntraints' and 'enableCustom' instead?

src/common/core/users/metarules.js
```diff
@@ -0,0 +1,311 @@
...
+                deferred.reject(new Error('Given nodePath does not exist "' + nodePath + '"!'));
+            } else {
+                deferred.resolve(node);
+            }
+        });
+
+        return deferred.promise;
+    }
+
+    function loadNodes(core, node, nodePaths) {
```
kecso: wouldn't it be enough to use Q.nfcall(core.loadByPath...), and then there is no need for the loadNode function?

test-karma/client/js/client.spec.js
```diff
@@ -184,16 +184,6 @@ describe('GME client', function () {
                 'getAspectTerritoryPattern'
             );
 
-            //addOn related API
-            expect(client).to.include.keys(
-                'validateProjectAsync',
```
kecso: shouldn't we check for the new function names?

config/config.default.js
pmeijer: My anticipation is that there will be more options for the core and not for the constraints.So adding a constraints group will clutter the config IMO.
Since the constraints are defined on core-nodes and the code for checking constraints is in common/core/users - core is not that awful.

src/common/core/users/metarules.js
pmeijer: There are two major differences.
1) If loadByPath returns err it is wrapped in an Error.
2) It checks if the node actually did exist or not.

Adding these to Q.nfcall(core.loadByPath...) will not make it more compact anymore.

test-karma/client/js/client.spec.js
pmeijer: Yes, that's correct. Adding it right now

pmeijer:
Alright I added the client API tests.

src/common/core/users/metarules.js
kecso: yes, you are right, I've missed that check.

config/config.default.js
kecso: agreed

PR #570, opened 2015-09-15
Fixes #566 css require paths should not depend on the document's path

### LINKED ISSUE #566: opening index.html directly screws styling
1. open localhost/index.html instead of localhost/
-> the styling of the page is odd (and there are some loading related error regarding some css files)

kecso:


lattmann:
I do not think that we want these changes, does it work under this #522 use case? This pull request, is the opposite of the #522.

pmeijer:
Tamas this fix works with /index.html and / if the config.client.appDir is the default. But not with a non-default.
On the other hand you're right about it not working without of this fix if /index.html is specified.
Suggested fix is the remove the index.html from document.location.pathname (if it exists).

kecso:
thanks for the comments.
Couldn't we use the appdir config option instead of removing the index.html (it seems a bit of hacking, but temporarily can suffice)?
I'll check if it is possible and let you know if I have a solution for both use-cases

kecso:
Patrik, I was able to make it work with some require rule changes (not the most beautiful, but works in all scenarios).
Please check it out!

pmeijer:
How does it work?

PR #574, opened 2015-09-16
Fix diagram designer tabs and droppable zone.

lattmann:


pmeijer:
On my chrome the whole canvas is no longer droppable if the window is big enough.

image

P.S. Snipping tool in Windows 10 has a Delay function :)

lattmann:
Good catch, thanks. Changes were made.

pmeijer:
Looks good, thank you!

PR #575, opened 2015-09-16
Revert "Fixes #378 create pointer after mouse over an object"

### LINKED ISSUE #378: Create pointer of an object menu doesn't show up after mouse is over of a pointer object
When an object that can be referenced to the editor, it can be dropped to the sheet. If the mouse is over to a pointer object in the sheet, the mouse icon turn into a link icon so the referenced object can be updated. However, after the mouse is once over a pointer object, it no longer allows creating a new pointer object.

pmeijer:
Reverts webgme/webgme#572

Just noticed that with these changes you can no longer click-and-drag an object on the canvas. Instead you need to click-drop-click-and-drag. If you click-and-drag the canvas will turn red.

### DIFF
--- a/src/client/js/Widgets/DiagramDesigner/DiagramDesignerWidget.Droppable.js
+++ b/src/client/js/Widgets/DiagramDesigner/DiagramDesignerWidget.Droppable.js
@@ -76,7 +76,6 @@ define([
 
         if (dragInfo) {
             this._doAcceptDroppable(this.onBackgroundDroppableAccept(event, dragInfo), true);
-            this._savedAcceptDroppable = this._acceptDroppable;
         } else {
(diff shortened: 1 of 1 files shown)


lattmann:
You need to select the object first that you like to drag. We may need to check how it works on operating systems and probably define all behavior. Would you like me to merge this?

pmeijer:
#571 sure seem like a more viable candidate for this issue. (Didn't pay attention to other UI changes.)
On windows you can drag and drop without selection. I'm afraid that, since it was working like this before, the change will irritate current users.

pmeijer:
This #378 was not the cause of the issue, but rather #571

lattmann:
I agree, let's iterate on it, and define the expected behavior next week.

PR #576, opened 2015-09-16
Simple in-memory storage for running plugins.

pmeijer:
Minor fixes for plugins in general.

### DIFF
--- a/src/server/api/index.js
+++ b/src/server/api/index.js
@@ -851,9 +851,9 @@ function createAPI(app, mountPath, middlewareOpts) {
         res.send({resultId: resultId});
     });
 
-    router.get('/plugins/results/:resultId', ensureAuthenticated, function (req, res) {
+    router.get('/plugins/:pluginId/results/:resultId', ensureAuthenticated, function (req, res) {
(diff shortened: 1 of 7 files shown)


pmeijer:
Creating pull-request, but open for discussion about the API and behavior.

src/server/api/index.js
```diff
@@ -735,26 +737,148 @@ function createAPI(app, mountPath, middlewareOpts) {
...
+            status: PLUGIN_CONSTANTS.RUNNING,
+            //timeoutId: will be added after plugin finished
+            //result: null,
+            //error: null
+        };
+
+        res.send({resultId: resultId});
+    });
+
+    router.get('/plugins/results/:resultId', ensureAuthenticated, function (req, res) {
```
lattmann: This would mean we cannot have a plugin that is called results, I think.

I suggest to put this functionality to `/plugins/:pluginId/results/:resultId`

lattmann:
It looks good. I made a minor comment.

src/server/api/index.js
pmeijer: Yes good point. I'll fix that.

Issue #563, opened 2015-09-14
We should not allow adding instance as descendant


kecso:
Currently we only forbid to add an instance to the node as a direct child (which means that parent !== base should stand for all nodes), but we not check for more distant relations.
The containment hierarchy and the inheritance chain of a node should never cross each other.

PR #583, opened 2015-09-18
Prohibit copy of FCO

kecso:
As some generic UI pieces depend on the fact, that there is only the FCO (and the ROOT) which doesn't have a base, we prohibit its copy.
This restriction however exsits only on the UI as in theory the core layer can handle multiple objects without base.

We also check for situations which cause unloadable nodes. More precisely when a node contains its base.
As not all situation can be avoided beforehand we remove such nodes during their load.
This issue is the opposite of #563
A possible scenario:
1. create typeA and typeB
2. create an instance of typeA (instA)
3. move typeA inside typeB
4. move typeB inside instA (at this point we are unable to prohibit the move as we not necessarily know about typeA and that it is the base of instA)
-> instA become unloadable

### DIFF
--- a/src/client/js/Panels/ModelEditor/ModelEditorControl.DiagramDesignerWidgetEventHandlers.js
+++ b/src/client/js/Panels/ModelEditor/ModelEditorControl.DiagramDesignerWidgetEventHandlers.js
@@ -427,8 +427,8 @@ define(['js/logger',
             };
 
         //check if FCO is among the items as it may change the outcome
-        for(i=0;i<items.length;i+=1){
-            if(GMEConcepts.isProjectFCO(items[i])){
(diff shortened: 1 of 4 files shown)


src/client/js/Panels/ModelEditor/ModelEditorControl.DiagramDesignerWidgetEventHandlers.js
```diff
@@ -425,6 +426,14 @@ define(['js/logger',
                 }
             };
 
+        //check if FCO is among the items as it may change the outcome
+        for(i=0;i<items.length;i+=1){
```
lattmann: style

PR #586, opened 2015-09-22
Add gmeConfig.server.extlibExcludes.

pmeijer:
Array of regexps that will return a 403 when accessed via the extlib routing rule.

By default all .pem files and the configuration files are added.


pmeijer:
Maybe it should be 404 here too? However 403 would make it easier for developers to see what is going on..

PR #588, opened 2015-09-23
Fixes #587 Inheritance connection upside down

### LINKED ISSUE #587: Inheritance connection upside down
Inheritance connections are backwards. That is, the triangle is connecting to the top and the other end is on the bottom.

inheritance_bug

I suspect this was introduced by #561

kecso:
inheritance connection 'base' end should connection to bottom of box by default
inheritance connection 'descendant' end should connection to top of box by default


brollb:
It works now - nice work!

PR #590, opened 2015-09-23
AddOns closes #579

### LINKED ISSUE #227: Switching between branches "quickly" kills addon.
To reproduce create a default project (i.e. with constraintAddon registered) with some nodes and branch of, say 4 times, and switch between these branches quickly.

This is the error:

```
ERROR - 16:42:23.775 [client] - stopAddOnError: addon
loadingLogManager.js:124
logMessageLogManager.js:135
logclient.js:420
callbackclient.js:392
stopAddOnclient.js:424
stopRunningAddOnsclient.js:1824 (anonymous function)
failsafe.js:289 (anonymous function)
client.js:698 (anonymous function)
socket.io.js:920
5.Socket.onacksocket.io.js:845
5.Socket.onpacketsocket.io.js:1197
8.module.exportssocket.io.js:1335
9.Emitter.emitsocket.io.js:395
3.Manager.ondecodedsocket.io.js:1197
8.module.exportssocket.io.js:1335
9.Emitter.emitsocket.io.js:5882
44.Decoder.addsocket.io.js:385
3.Manager.ondatasocket.io.js:1197
8.module.exportssocket.io.js:1335
9.Emitter.emitsocket.io.js:1953
13.Socket.onPacketsocket.io.js:1771 (anonymous function)
socket.io.js:1335
9.Emitter.emitsocket.io.js:2378 14.
Transport.onPacketsocket.io.js:2370 14.
Transport.onDatasocket.io.js:3433 19.
WS.addEventListeners.ws.onmessage
```

### LINKED ISSUE #467: AddOn Generator
AddOns, like plugins, should have an generator plugin for creating the boilerplate files.

### LINKED ISSUE #579: Major AddOn related improvements.
### AddOnAPI
- [x] AddOn.onUpdate(rootNode, commitObj, callback) - called when changes are made in the branch where addOn is running. AddOns are allowed to modify the core/rootNode - but not persist by themselves.
- [ ] AddOn.query(commitHash, queryParams, callback) - Invoked from the clients. AddOns are not suppose to make any modifications to the model - just query it. QueryResult will be similar (if not the same as PluginResult) - **Will be in later versions, for now use Plugins.**
- [x] AddOn.getQueryParameterStructure (static) equivalent of plugin.getConfigurationStructure.
- [x] AddOn.getDefaultQueryParameters (static) equivalent of plugin.getDefaultConfig
- [x] getName, getVersion, getDescription, etc. Same as Plugin.

### AddOnManager (Node)
- [x] AddOnManager running in node workers.
- [x] Keeps one project open and can have multiple branches open.
- [x] Should open a branch and instantiate AddOns.
- [x] Each branch can have multiple AddOns running (reads from the model).
- [x] Persists changes to the rootNode made by AddOns.

### WebSocket/ServerWorkerManager
- [ ] Maps each connected client (based on their opened project) to a worker.
- [ ] Needs to start and kill workers based on number of connected clients - **In this version only a single connected-worker is running.**

## Open Questions
When the AddOnManager opens the storage. Which user should it open it with? All users viewing the model with addOns running are guaranteed to have the correct authorization level. (AddOns can only make changes as a response to changes made by a client). 
Suggestion is to have a dedicated addOn user with full rights to all projects. That way it is also easier to distinguish it from 'clients' when mapping users to addOns.
**In this version the sessionId of the first connecting client to a project is used for all branches in that project.**

## For the Future
AddOns should be able to send notifications to clients.

pmeijer:
Above add-on specifics this branch contains:

- Notifications to clients when other sockets join the same branch-room.
- Clear intervals/timeouts in ExecutorServer on stop.
- Disconnect "login" sockets in common-storage tests.
- Minor tweaks to tests.

Closes #227
Closes #467


src/addon/addonmanager.js
```diff
@@ -6,134 +6,281 @@
...
-        if (!addOn) {
-            deferred.reject(new Error('The addOn is not running'));
-        } else {
-            addOn.query(parameters, function (err, message) {
-                if (err) {
-                    deferred.reject(err instanceof Error ? err : new Error(err));
+        if (monitor) {
+            monitor.connectionCnt += 1;
+            logger.debug('monitorBranch [' + branchName + '] - connection counter [prev, now]',
+                monitor.connectionCnt - 1, monitor.connectionCnt);
```
kecso: is it worth printing out n, n+1 all the time? (I mean if that is the only option, it seems a bit redundant)

src/addon/addonmanager.js
```diff
@@ -6,134 +6,281 @@
...
+                    // We cannot simply remove the monitor before it has closed the branch,
+                    // therefore we need to register on the stop.promise and start a new monitor
+                    // once it has resolved.
+                    //
+                    // (For bookkeeping of the addOnManagers it is not the same since they create a new storage.)
+                    self.inStoppedAndStarted[branchName] = true;
+                    logger.debug('monitorBranch [' + branchName + '] - inStoppedAndStarted!');
+                    monitor.instance.stop()
+                        .then(function () {
+                            delete self.inStoppedAndStarted[branchName];
```
kecso: does this inStoppedandStarted has any real effect?
- it seems that it can cause confusion
- what if multiple watcher arrives during the stop period? (right now it seems we will have an error instead of starting all...)
- if only the 'first' callback would try to start the new monitor and all the others would just connect, that could solve the confusion (and would give meaning to the flag as well)

src/addon/addonmanager.js
```diff
@@ -6,134 +6,281 @@
...
-        }
-
-        addOn = runningAddOns[addOnName];
+        if (!monitor) {
+            logger.debug('No monitor [' + branchName + '] assuming this is the monitor itself disconnecting');
+            deferred.resolve(-1);
+        } else {
+            monitor.connectionCnt -= 1;
+            logger.debug('unMonitorBranch [' + branchName + '] - connection counter [prev, now]',
+                monitor.connectionCnt + 1, monitor.connectionCnt);
```
kecso: same as above, the two number seems redundant as we always decrease with one...

src/addon/branchmonitor.js
```diff
@@ -0,0 +1,301 @@
...
+                        startDeferred.reject(err);
+                        return;
+                    }
+                    self.branchIsOpen = true;
+                    startDeferred.resolve();
+                }
+            );
+        }
+
+        return startDeferred.promise.nodeify(callback);
```
kecso: does it work if the promise already resolved? (meaning that somehow a monitor.start arrives after the initial request has been handled)

src/server/storage/websocket.js
```diff
@@ -11,7 +11,6 @@ var io = require('socket.io'),
     COOKIE = require('cookie-parser'),
     URL = requireJS('common/util/url'),
     CONSTANTS = requireJS('common/storage/constants'),
-    ROOM_DIV = CONSTANTS.ROOM_DIVIDER, // TODO: Add prefixes
     DATABASE_ROOM = CONSTANTS.DATABASE_ROOM;
```
kecso: sidenote: couldn't we remove the DATABASE_ROOM contant as well?

src/addon/addonmanager.js
pmeijer: agree, it is pretty silly

src/addon/branchmonitor.js
pmeijer: Yes, that is happening all the time.
In addOnManager.monitorBranch (invoked on each socket room join) each invocation reuses this promise. It's a neat feature of the q library. Kevin suggested that for mongo connections in #482.

src/server/storage/websocket.js
pmeijer: Sure good point.

src/addon/addonmanager.js
pmeijer: answer same as above ;)

src/addon/addonmanager.js
pmeijer: It does cover for the case where one new socket connects in that small time window. There is a test case for it.
But yes, multiple incoming ones should register on the stop-promise too and when that is resolved register on the start-promise. (With multiple registrations on the promise - the will resolve/reject in the order they were registered).

pmeijer: P.S. That is a really good catch. Please see my addition.

PR #591, opened 2015-09-23
Fix: fall back on rootNode when config.defaultProject.node is not given.

pmeijer:


src/client/js/WebGME.js
```diff
@@ -89,8 +89,10 @@ define([
             // URL query has higher priority than the config.
             if ((initialThingsToDo.projectToLoad || initialThingsToDo.createNewProject) === false) {
                 initialThingsToDo.projectToLoad = gmeConfig.client.defaultProject.name;
-                initialThingsToDo.branchToLoad = initialThingsToDo.branchToLoad || gmeConfig.client.defaultProject.branch;
-                initialThingsToDo.objectToLoad = initialThingsToDo.objectToLoad || gmeConfig.client.defaultProject.node;
+                initialThingsToDo.branchToLoad = initialThingsToDo.branchToLoad ||
+                    gmeConfig.client.defaultProject.branch;
+                initialThingsToDo.objectToLoad = initialThingsToDo.objectToLoad ||
+                    gmeConfig.client.defaultProject.node || initialThingsToDo.objectToLoad; // i.e. the root-node.
```
kecso: I think it is still not right.
First of all, if you have anything in the config, then it will override the ROOT path.
Secondly, the intialThingsToDo.objectToLoad already 'defaulted' the node path to ROOT path, which makes the whole config option meaningless in this form.
-> we either loose this config option, or we not default the param from the URL, and then the expression could be URL || config || ROOT

pmeijer: But the Url should have priority. In the url parser the default return is '', which is falsy. The issue was when the config had null (the default), now if that is falsy it again falls back on the root node.

kecso: exactly.
I was not saying that it not fixed the original issue. I was pointing out, that it still not the best we can come up with.
so if we think about the scenario I wrote, then we can simplify the question on whether the config has higher priority than the url parameters.
I think the url parameters has the highest priority which is why I proposed the two alternative options to fix the issue as well as get rid of another.

pmeijer: The config only applies if the url did not specify a project.

pmeijer:
Wait with pr

src/client/js/WebGME.js
pmeijer: It works as expected. You need to try it if you doubt it.

PR #598, opened 2015-09-25
Adds a cached meta layer to core

### LINKED ISSUE #377: Part Browser is not updated correctly when switching between branches.
When switching between branches within the same project, sometimes only the FCO shows up in the part browser for the root.

When switching between projects it works.

kecso:
To harmonize META related queries we created a new layer for the core, that loads all the META nodes of the project, to be able to answer more complex questions. This layer now understands the concept of METAAspectSet, so it is not a solely UI concept anymore.
We removed the METAAspectHelper, as its functionality is not completely covered by the core.
The context menu of the TreeBrowser now uses the new functions provided by the client (that uses the new core functionality)
The part browser has also been revised to use the new client functions. It has a much simpler state-machine now.
Eventually GMEConcepts would be removable as well.
As a result of the refactor, META name collision now handled and will not cause issues (every item can be instantiated, though in some cases it will be hard do distinct among META objects with the same name). A notification will still appear to let the user know, that there is something wrong with the META of the project.

Fixes #377


src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
```diff
@@ -346,7 +345,7 @@ define(['js/logger',
                 //    "icon": false
                 //};
 
-                if (GMEConcepts.getCrosscuts(nodeId).length > 0) {
+                if (nodeObj.getCrosscutsInfo().length > 0) {
```
pmeijer: nodeObj needs to be checked for null to have the same behavior as before.

src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
```diff
@@ -628,66 +627,30 @@ define(['js/logger',
...
-        }
-
-        children.sort(function (a, b) {
-            if (a.Title.toLowerCase() < b.Title.toLowerCase()) {
-                return -1;
-            } else {
-                return 1;
+        var types = [],
+            node = this._client.getNode(nodeId),
+            validChildrenInfo = node.getValidChildrenTypesDetailed(),
```
pmeijer: node needs to be checked for null here too.

src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
```diff
@@ -628,66 +627,30 @@ define(['js/logger',
...
-
-        if (GMEConcepts.canCreateChild(nodeId, childId)) {
-            var params = {parentId: nodeId};
-            params[childId] = {registry: {}};
-            params[childId].registry[REGISTRY_KEYS.POSITION] = {x: 100, y: 100};
-            client.createChildren(params);
-        } else {
-            logger.warn('Can not create child instance of \'' + childId + '\', in parent object: \'' + nodeId + '\'');
-        }
+        var client = this._client;
```
pmeijer: Here too

src/common/core/cachedmetacore.js
```diff
@@ -0,0 +1,363 @@
...
+                return validPtrNames.indexOf('dst') !== -1 && validPtrNames.indexOf('src') !== -1;
+            };
+
+            function sensitiveFilter(validNodes) {
+                var i;
+
+                i = validNodes.length;
+                while (i--) {
+                    if (core.isConnection(validNodes[i]) || core.isAbstract(validNodes[i])) {
+                        validNodes.splice(i, 1);
```
pmeijer: Why use splice instead of pop? http://jsperf.com/pop-vs-splice

src/common/core/cachedmetacore.js
```diff
@@ -0,0 +1,363 @@
...
+                if (setName === META_SET_NAME && core.getPath(node) === core.getPath(root)) {
+                    delete root.metaElements[memberPath];
+                }
+            };
+
+            core.deleteNode = function (node, technical) {
+                var root = core.getRoot(node);
+                if (root.metaElements[core.getPath(node)]) {
+                    delete root.metaElements[core.getPath(node)];
+                }
```
pmeijer: For the future: the path should be cached. Ideally in the node, who knows how many getPaths are being called through all the core layers?

src/common/core/cachedmetacore.js
```diff
@@ -0,0 +1,363 @@
...
+                    rules,
+                    inAspect,
+                    temp;
+
+                rules = oldcore.getChildrenMeta(node) || {};
+
+                for (i = 0; i < keys.length; i += 1) {
+                    temp = metaNodes[keys[i]];
+                    while (temp) {
+                        if (rules[oldcore.getPath(temp)]) {
```
pmeijer: is oldcore.getPath any different from core.getPath?

src/common/core/cachedmetacore.js
```diff
@@ -0,0 +1,363 @@
...
+                return validNodes;
+            };
+
+            //parameters
+            // node - the node in question
+            // name - the name of the set
+            // members - the current members of the set, so that multiplicity can be checked
+            // sensitive - if true the function do not return the connection and abstract types
+            // multiplicity - if true the function filters out possibilities that fail multiplicity check
+            core.getValidSetElementsMetaNodes = function (parameters) {
```
pmeijer: Why not JsDoc here?
/** + Enter - in webstorm

src/common/core/cachedmetacore.js
```diff
@@ -0,0 +1,363 @@
...
+                    i = validNodes.length;
+                    keys = Object.keys(typeCounters);
+                    while (i--) {
+                        for (j = 0; j < keys.length; j += 1) {
+                            if (rules[keys[j]].max &&
+                                rules[keys[j]].max > -1 &&
+                                rules[keys[j]].max <= typeCounters[keys[j]] &&
+                                oldcore.isTypeOf(validNodes[i], metaNodes[keys[j]])) {
+                                validNodes.splice(i, 1);
+                                break;
```
pmeijer: again why splice?

src/common/core/cachedmetacore.js
kecso: we talked that the core will have a separate place for the documentation

src/common/core/cachedmetacore.js
kecso: nope, I just thought it is safer as it may be overridden in this layer, and then it can give back something else (though it is not too likely)

src/common/core/cachedmetacore.js
kecso: Path can change, and these function calls should not take too muuch time, but we defenetly should check it.

src/common/core/cachedmetacore.js
kecso: pop always pops the last element, however this is not the case here

src/common/core/cachedmetacore.js
```diff
@@ -0,0 +1,363 @@
...
+                var root = core.getRoot(node);
+                oldcore.addMember(node, setName, member);
+
+                //check if our cache needs to be updated
+                if (setName === META_SET_NAME && core.getPath(node) === core.getPath(root)) {
+                    root.metaElements[core.getPath(member)] = member;
+                }
+            };
+
+            core.delMember = function (node, setName, memberPath) {
```
pmeijer: In the top-layer I would add a core.removeMember
createNode - deleteNode
addMember - removeMember (not delMemeber)

src/common/core/cachedmetacore.js
kecso: same as before.
it is possible that you remove a middle element

src/common/core/cachedmetacore.js
```diff
@@ -0,0 +1,363 @@
...
+                logger = options.logger.fork('meta');
+            for (var key in oldcore) {
+                core[key] = oldcore[key];
+            }
+            logger.debug('initialized');
+
+            function loadMetaSet(root) {
+                var paths = oldcore.getMemberPaths(root, META_SET_NAME),
+                    i,
+                    metaElements = [];
```
pmeijer: Minor thing, why metaElements and not metaNodes?
getAllMetaNodes returns with metaElements..

src/common/core/cachedmetacore.js
kecso: what do you mean?
adding a new name or renaming the public API?

src/common/core/cachedmetacore.js
kecso: good question.
I will rename it.

src/common/core/cachedmetacore.js
pmeijer: true

test/common/core/core.intrapersist.js
```diff
@@ -95,7 +112,7 @@ describe('core.intrapersist', function () {
                 done();
             });
         });
-        it('loads all the nodes for the test', function (done) {
+        it.skip('loads all the nodes for the test', function (done) {
             loadNodes([e1NodePath, e1NodePrimePath, s1NodePath, s1NodePrimePath], function (err, n) {
                 if (err) {
                     return done(err);
```
pmeijer: What is going on here?

test/common/core/cachedmetacore.spec.js
```diff
@@ -0,0 +1,372 @@
...
+        for (i = 0; i < keys.length; i += 1) {
+            baseNodes[keys[i]] = null;
+        }
+
+        //we load all the nodes that we wish to work with
+        Q.nfcall(core.loadRoot, baseRootHash)
+            .then(function (root) {
+                rootNode = root;
+                baseNodes[''] = root;
+                return Q.allSettled([
```
pmeijer: Could use Q.allDone

test/common/core/cachedmetacore.spec.js
```diff
@@ -0,0 +1,372 @@
...
+    it('should return all meta nodes', function () {
+        var metaPaths = ['/1',
+            '/367050797',
+            '/367050797/1776478501',
+            '/367050797/625420143',
+            '/367050797/1626677559',
+            '/367050797/355480347'
+        ];
+
+        expect(Object.keys(core.getAllMetaNodes(baseNodes['/1924875415']) || {})).to.have.members(metaPaths);
```
pmeijer: Wouldn't you typically pass the rootNOde to getAllMetaNodes?

test/common/core/cachedmetacore.spec.js
kecso: sure

PR #611, opened 2015-10-01
Fixes #565 Re-defining a set should not restore its old members

### LINKED ISSUE #565: Re-defining a set should not restore its old members
1. define a set
2. add some members to it
3. remove the meta-rule
4. define the set again
-> the members are still there (though the set was deleted when the rule was removed)

kecso:
Now set's are defined on the node's level even in case of inherited set extension, so after the removal of the base type's set it can be visible and extension members can be removed so after re-definition they will not appear.
However if the set in the base is re-defined without touching the instance's set, its members will remain there so become visible (because its data haven't been touched...).

### DIFF
--- a/src/common/core/setcore.js
+++ b/src/common/core/setcore.js
@@ -16,12 +16,19 @@ define(['common/util/assert'], function (ASSERT) {
         ASSERT(typeof options.globConf === 'object');
         ASSERT(typeof options.logger !== 'undefined');
 
-        var logger = options.logger.fork('setcore');
+        var logger = options.logger.fork('setcore'),
+            setCore = {};
+        for (var i in innerCore) {
+            setCore[i] = innerCore[i];
+        }
+        logger.debug('initialized');
+
         //help functions
-        var setModified = function (node) {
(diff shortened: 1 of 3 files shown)


src/common/core/setcore.js
```diff
@@ -69,9 +78,9 @@ define(['common/util/assert'], function (ASSERT) {
                 relid = Math.floor(Math.random() * MAX_RELID);
             } while (existingRelIds.indexOf(relid) !== -1);
             return '' + relid;
-        };
+        }
 
-        var harmonizeMemberData = function (node, setName) {
+        function harmonizeMemberData(node, setName) {
```
pmeijer: Looks good in general, just a minor note for the future.
This function has 12 variables, 7 levels of indentation and three levels of nested loops and it's a recursive function. Do consider refactoring out the inner most loop into a helper function (ideally outside of the function scope).

kecso: Thanks, noted!

I wasn't checking the function (just removed the var), but next time I am around, I will definitely refactor it.

PR #624, opened 2015-10-09
Added custom layout support. Fixes #606

### LINKED ISSUE #606: Custom Layout support
There looks to be some support for custom layouts ([here] (https://github.com/webgme/webgme/tree/master/src/client/js/Layouts)) and should only require minor changes ([here](https://github.com/webgme/webgme/blob/master/src/client/js/LayoutManager/LayoutManager.js#L12) and [here](https://github.com/webgme/webgme/blob/master/src/client/js/LayoutManager/LayoutManager.js#L45)) to support custom layouts without entirely rebuilding the ui from scratch.

These changes would also require some changes to the config structure.

brollb:
Added custom layout support. Fixes #606

Custom layout explorations. WIP #606

Added custom layout support. Fixes #606

WIP #606

Added custom layout. Fixes #606

Added custom layout support. Fixes #606

Removed extra files #606

Removed extra files. #606


config/config.default.js
```diff
@@ -190,7 +190,9 @@ var path = require('path'),
         visualization: {
             decoratorPaths: [path.join(__dirname, '../src/client/decorators')],
             visualizerDescriptors: [path.join(__dirname, '../src/client/js/Visualizers.json')]
-        }
+        },
+
+        layout: 'js/Layouts/DefaultLayout/DefaultLayout'
```
pmeijer: Does this mean we need to add custom requirejs paths for layouts? It's probably better do handle them the same way as Plugins and AddOns using basePaths.

pmeijer: Actually under visualization would be even better.

webgme.js
```diff
@@ -46,8 +47,9 @@ function addToRequireJsPaths(gmeConfig) {
                 try {
                     items = fs.readdirSync(basepaths[j]);
                     if (items.indexOf(componentNames[i]) !== -1) {
-                        componentPaths[componentName + '/' + componentNames[i]] = path.relative(requireJsBase,
-                            path.resolve(basepaths[j]));
+                        componentPath = path.relative(requireJsBase, path.resolve(basepaths[j]))
+                            .replace(/^client\/js/, 'js');
+                        componentPaths[componentName + '/' + componentNames[i]] = componentPath;
```
pmeijer: Do we really need the replace part here? It can potentially mess up custom paths.

brollb: This was added because currently all components using this function (the components with `basePaths` in the config) have the same path on the filesystem as their path hosted on the server (eg, `plugin`, `common` as both are in `src/`).

However, when we are adding base paths for `layout`, the path on the server doesn't match the path on the file system. That is, if a component has `basePaths` set in the config but exists in `src/client/js` on the filesystem, it's path on the server is just `js/rest/of/path`.

Anyway, this should not cause problems with any other components unless we decide to add a `client/js` on the server which does not correspond to the `client/js` on the file system.

PR #628, opened 2015-10-12
loadPaths function

kecso:
By adding a loadPaths function to the storages we are able to pre-load multiple objects from the server saving many roundtrips.
Right now, this functionality is used when we load the MetaNodes and when we loadByPath.
It also implemented inside the loadCollection function, but that is rarely used, and whenever we will implement the loadMembers, that function could benefit from it as well.

### DIFF
--- a/src/common/storage/project/cache.js
+++ b/src/common/storage/project/cache.js
@@ -95,6 +95,99 @@ define(['common/util/assert', 'common/storage/constants'], function (ASSERT, CON
             callback(null, obj);
         };
 
+        this.loadPaths = function (rootKey, paths, callback) {
+            ASSERT(typeof rootKey === 'string' && paths instanceof Array && typeof callback === 'function');
+            logger.debug('loadPaths', {metadata: {rootKey: rootKey, paths: paths}});
+
+            function getFromCache(hash) {
+                var obj = cache[hash],
+                    commitId;
+
+                if (typeof obj === undefined) {
+                    obj = backup[hash];
+
+                    if (typeof obj === undefined) {
+                        for (commitId in self.queuedPersists) {
+                            if (self.queuedPersists.hasOwnProperty(commitId) && self.queuedPersists[commitId][key]) {
+                                obj = self.queuedPersists[commitId][key];
+                                break;
+                            }
+                        }
+                    }
+                }
+
+                return obj;
+            }
+
+            var objects = {},
+                excludes = [],
+                pathArray,
+                fullyCovered,
+                obj,
+                rootObj = getFromCache(rootKey),
+                key, i = paths.length,
+                j;
+
+            if (typeof rootObj !== undefined) {
+                excludes.push(rootKey);
+                objects[rootKey] = rootObj;
+
+                while (i--) {
+                    fullyCovered = true;
+                    pathArray = paths[i].split('/');
+                    obj = rootObj;
+                    for (j = 0; j < pathArray.length; j += 1) {
+                        key = obj[pathArray[j]];
+                        if (key) {
+                            obj = getFromCache(key);
+                            if (typeof obj !== undefined) {
+                                excludes.push(key);
+                                objects[key] = obj;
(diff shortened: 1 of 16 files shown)


src/common/storage/project/cache.js
```diff
@@ -95,6 +95,104 @@ define(['common/util/assert', 'common/storage/constants'], function (ASSERT, CON
...
+                            }
+                        }
+                    } else {
+                        logger.warn('loadingPaths failed', err || new Error('no bject arrived from server'));
+                    }
+                    callback(err);
+                });
+            } else {
+                callback(null);
+            }
```
pmeijer: Why return with null if the root object did not exist in the cache?

src/common/storage/project/cache.js
```diff
@@ -95,6 +95,104 @@ define(['common/util/assert', 'common/storage/constants'], function (ASSERT, CON
...
+                    }
+
+                    if (fullyCovered) {
+                        paths.splice(i, 1);
+                    }
+                }
+
+                //now we checked the cache, so if something remains we ask it from the server
+                if (paths.length === 0) {
+                    return callback(null);
```
pmeijer: break in two lines

src/server/storage/memory.js
```diff
@@ -100,6 +101,69 @@ function Memory(mainLogger, gmeConfig) {
             return deferred.promise.nodeify(callback);
         };
 
+        this.loadPaths = function (rootHash, paths, excludes, callback) {
```
pmeijer: This pretty complex piece of code is copied here and in Mongo. Couldn't it be implemented in the storage instead?

src/common/storage/project/cache.js
```diff
@@ -95,6 +95,104 @@ define(['common/util/assert', 'common/storage/constants'], function (ASSERT, CON
...
+                        keys = Object.keys(objects);
+                        for (i = 0; i < keys.length; i += 1) {
+                            if (!cache[keys[i]]) {
+                                cacheInsert(keys[i], objects[keys[i]]);
+                            }
+                        }
+                    } else {
+                        logger.warn('loadingPaths failed', err || new Error('no bject arrived from server'));
+                    }
+                    callback(err);
```
pmeijer: Any potential issues if the cache/backup has been cleared of the 'excludes' when we get to this point?

test/server/storage/mongo.spec.js
```diff
@@ -1248,4 +1248,81 @@ describe('Mongo storage', function () {
...
+            Q.nfcall(project.loadPaths, '#r1000000000000000000', [''], [])
+                .then(function (objects) {
+                    expect(Object.keys(objects)).to.have.members(['#r1000000000000000000']);
+                    done();
+                })
+                .catch(done);
+        });
+
+    });
 });
```
pmeijer: Would be nice with some tests testing the full path from the core.

src/server/storage/memory.js
kecso: My first intent was to put this function completely into the storage, but it turned out that if I want it to function for all types of users (bin scripts, tests, worker, webclient) there is no single point where the function could be implemented.
If we fine with the pre-loading not working for every type of users (mostly the webclient could gain from it), then I can put the functionality into the storage and have an empty function for the others.

src/common/storage/project/cache.js
kecso: as we store a reference for the objects we found in the cache/backup/commitQueue we should be able to re-insert them if needed at this point.

pmeijer: Ah true, you insert the objects from "objects"

src/server/storage/memory.js
pmeijer: Ok, no it's definitely better if we can keep the feature the same across the different implementations.
In general though, we should perhaps consider having either a base class or a set of utility functions for the "data-store" classes.

PR #631, opened 2015-10-15
Feature/enhanced meta attribute

kecso:
Numerical attributes now can have inclusive minimum and maximum values.
String attributes can have a regular expression defined to match with.

### DIFF
--- a/src/common/core/metacore.js
+++ b/src/common/core/metacore.js
@@ -147,6 +147,7 @@ define([
                 return true;
             }
             var pointerMetaNode = getMetaPointerNode(source, name);
+            
             if (pointerMetaNode) {
                 var validTargetTypePaths = core.getMemberPaths(pointerMetaNode, 'items') || [];
                 while (node) {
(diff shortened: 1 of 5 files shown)


src/common/core/metacore.js
```diff
@@ -190,12 +200,20 @@ define([
                     break;
                 case 'integer':
                     if (!isNaN(parseInt(value)) && parseFloat(value) === parseInt(value)) {
-                        return true;
+                        if ((!meta.min || (meta.min && value >= meta.min)) &&
+                            (!meta.max || (meta.max && value <= meta.max))) {
```
pmeijer: What if meta.min or meta.max is zero?
And why the extra check in the or clause?

lattmann: What is the type of the value here? Do we need `parseInt(value)` again for the comparison?

kecso: you are right, the condition should change, as the 0 is falsy

lattmann:
Thanks for the changes.

PR #636, opened 2015-10-16
Feature/nodeless inquiry

kecso:
To speed up inquiries on core we implemented the functions in a manner when no sub-gmeNode objects are created during inquiry (the way of manipulating have not changed).
In the process we reviewed the setcore and make some changes:
- getOwnMemberPaths now available on the core API to get those members of a set, that has some information on the node's level (so not completely inherited)
- isFullyOverriddenMember now available on core API to check if a member have been added to the set of the node, though it was inherited as well
- a new optional flag 'overridden' is now used in the export format when describing the set members, to capture if a member was completely overridden on the node's inheritance level (so added to the set directly on the node)

To measure the speed increase we simply exported a relatively bigger model (189Kb in export size):
- with the current master branch, the speed was varying from 1605ms to 2038ms
- with the nodeless_inquiry addition, the speed was reduced to the range of 756ms to 1033ms
This is roughly a ***50%*** increase or we can say it is ***twice*** as fast as the original one. We have to mention that this increase may not effect (or not effect to this extent) all UI operation.

### DIFF
--- a/src/common/core/setcore.js
+++ b/src/common/core/setcore.js
@@ -8,8 +8,11 @@
 define(['common/util/assert'], function (ASSERT) {
     'use strict';
 
-    var SETS_ID = '_sets';
-    var REL_ID = 'member';
+    var SETS_ID = '_sets',
+        REL_ID = 'member',
+        OVERLAYS = 'ovr',
+        ATTRIBUTES = 'atr',
+        REGISTRY = 'reg';
 
     function SetCore(innerCore, options) {
         ASSERT(typeof options === 'object');
@@ -26,6 +29,13 @@ define(['common/util/assert'], function (ASSERT) {
         //help functions
         function setModified(node) {
             innerCore.setRegistry(node, '_sets_', (innerCore.getRegistry(node, '_sets_') || 0) + 1);
+        }
+
+        function relIdSelector(key) {
+            if (key.indexOf('_') !== 0 && key !== REGISTRY && key !== ATTRIBUTES && key !== OVERLAYS) {
+                return true;
+            }
+            return false;
         }
 
         function getMemberPath(node, setElementNode) {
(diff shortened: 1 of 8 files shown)


lattmann:
What is the speed up by this change? Where are the measurements?

kecso:
Thanks for the comment!
I've updated the PR with a quick measurement on the export (I've exported one version of an ADM project). Please let me know if you need any more details.

src/common/core/setcore.js
```diff
@@ -80,65 +100,29 @@ define(['common/util/assert'], function (ASSERT) {
...
-                                }
-                                innerCore.deleteNode(innerCore.getChild(setNode, ownMembers[i]), true);
-                            }
+        function collectSetNames(node) {
+            var sets = [],
+                setsInfo,
+                keys, i;
+            do {
+                setsInfo = setCore.getProperty(node, SETS_ID);
+                if (setsInfo && setsInfo['ovr'] && setsInfo['ovr']['']) {
```
pmeijer: Use OVERLAYS and consider moving these to a separate core constants file.

src/common/core/setcore.js
```diff
@@ -209,53 +252,32 @@ define(['common/util/assert'], function (ASSERT) {
...
-            harmonizeMemberData(node, setName);
-            var memberRelId = getMemberRelId(node, setName, memberPath);
-            if (memberRelId) {
-                var memberNode = innerCore.getChild(
-                    innerCore.getChild(innerCore.getChild(node, SETS_ID), setName), memberRelId);
-
-                return innerCore.getAttributeNames(memberNode);
-            }
-            return [];
+            return collectPropertyNames(node, setName, memberPath, 'atr');
```
pmeijer: Use constant for attributes.

PR #638, opened 2015-10-19
Added panelPaths to config. Fixes #625

### LINKED ISSUE #625: Panel basePaths
Panels should be hosted on a `panel/` endpoint like plugins and layouts.

The lack of basePaths currently prevents any custom panels to be loaded by default in a layout because of [this](https://github.com/webgme/webgme/blob/master/src/client/js/LayoutManager/LayoutManager.js#L70). That is, it checks in the default location in the webgme project itself.

brollb:
Updated panel basePaths to be consistent with plugins/addons. #625

Note: Check out the comments in `webgme.js` and `src/server/standalone.js`

### DIFF
--- a/src/server/standalone.js
+++ b/src/server/standalone.js
@@ -386,6 +386,51 @@ function StandAloneServer(gmeConfig) {
         }
     }
 
+    function getGoodExtraAssetRouteFor(component, basePaths) {
+        // Check for good extra asset
+        return function(req, res) {
+            res.sendFile(Path.join(__baseDir, req.path), function (err) {
+                if (err && err.code !== 'ECONNRESET') {
+                    //this means that it is probably plugin/pluginName or plugin/pluginName/relativePath format
+                    // so we try to look for those in our config
+                    //first we check if we have the plugin registered in our config
+                    var urlArray = req.url.split('/'),
+                        pluginName = urlArray[2] || null,
+                        basePath,
+                        baseAndPathExist,
+                        relPath;
+
+                    relPath = getRelPathFromUrlArray(urlArray);
+                    basePath = getBasePathByName(pluginName, basePaths);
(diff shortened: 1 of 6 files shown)


src/server/standalone.js
```diff
@@ -386,6 +386,51 @@ function StandAloneServer(gmeConfig) {
...
+            relPath += '.js';
+        }
+        return relPath;
+    }
+
+    /**
+     * Unlike `getGoodExtraAssetRouteFor`, `getRouteFor` does not assume that the 
+     * resource hosts a main file which has the same structure as the parent directory.
+     * That is, there are examples of panels (such as SplitPanel) in which the
+     * main file does not adhere to the format "NAME/NAME+'Panel'"
```
pmeijer: If there is no common rule for that format, how can it know which file to require?

webgme.js
```diff
@@ -80,6 +80,28 @@ function addToRequireJsPaths(gmeConfig) {
...
     addFromBasePath(gmeConfig.visualization.layout.basePaths, 'layout');
+
+    // Panels are structured a little differently and may not have
+    // PANEL_NAME/PANEL_NAME.js like plugins, addons or layouts
+    var paths = gmeConfig.visualization.panelPaths,
+        panelPaths = {},
+        itemPath,
+        items;
+
+    for (var i = paths.length; i--;) {
```
pmeijer: Could you refactor this into a function similar to addFromBasePath?
A minor style thing is to put the variable declarations on the top of the scope.

src/server/standalone.js
brollb: The file is specified when you are requiring it. In other words, if I want to use the SplitPanel, I would require `panel/SplitPanel/SplitPanel` and if I wanted to reuse the PointerWidget (from the PropertyEditor), I would require `panel/PropertyEditor/PointerWidget`

Does that answer your question?

PR #639, opened 2015-10-19
Move loadPaths and getCommonAncestor to storage.

pmeijer:
This removes duplicate and complex code from the database adapters.
cachInsert (private method in cache) does no longer throw exceptions on existing objects - instead it returns true or false.

This PR also contains changes to loadPaths behavior #628


src/common/storage/project/cache.js
```diff
@@ -95,102 +120,98 @@ define(['common/util/assert', 'common/storage/constants'], function (ASSERT, CON
...
-                                break;
-                            }
+
+                key;
+
+            if (!rootKey || !rootObj) {
+                logger.debug('rootKey empty or rootObject was not loaded in cache, rootKey:', rootKey);
+                callback(null);
+                return;
+            }
```
kecso: could we correct the behavior so that if even the root object were not found we still send out the request to the server with an empty excludes array? (means that we put the whole internal loading to the else branch of this if statement)

test/server/storage/loadPaths.json
```diff
@@ -0,0 +1,397 @@
+{
```
kecso: according to our test structures, this resource file should be in directory test/server/storage/safestorage

src/common/storage/project/cache.js
pmeijer: Yes, I will return on faulty rootKey though (without it the server could not figure out where to start).

test/server/storage/loadPaths.json
pmeijer: Correct, I will fix it.

Issue #641, opened 2015-10-21
AddOn Generator Tests


brollb:
The AddOnGenerator has no tests

brollb:
Fixed in https://github.com/webgme/webgme/commit/9b750ce88f289e007828e04e46add71ac957d3b7

PR #642, opened 2015-10-21
Added test and fixed AddOnBase path. Fixes #640, #641

### LINKED ISSUE #640: Generated addOns have incorrect AddOnBase path
In the dependency array, the generated addon requires `addOn/AddOnBase` when it should require `addon/AddOnBase`

brollb:


pmeijer:
Thanks a lot Brian.

PR #643, opened 2015-10-21
Filter in ProjectsDialog applies to names.

pmeijer:
Better descriptions for options when forked.


pmeijer:
Will rebase this branch..

PR #646, opened 2015-10-22
Allow REST access to raw node data from REST.

pmeijer:
`/api/projects/[ownerId]/[projectName]/commits/[commitHash]/tree/path/to/object`

`/api/projects/[ownerId]/[projectName]/branches/master/tree/path/to/object`

Some tweaks and improvements to the storage, esp. w.r.t. loadPaths.

### DIFF
--- a/src/server/api/index.js
+++ b/src/server/api/index.js
@@ -529,6 +529,43 @@ function createAPI(app, mountPath, middlewareOpts) {
 
     // PROJECTS
 
+    function loadNodePathByCommitHash(userId, projectId, commitHash, path) {
+        var getCommitParams = {
+            username: userId,
+            projectId: projectId,
+            number: 1,
+            before: commitHash
+        };
+
+        return safeStorage.getCommits(getCommitParams)
+            .then(function (commits) {
+                var loadPathsParams = {
+                    projectId: projectId,
+                    username: userId,
+                    pathsInfo: [
+                        {
+                            parentHash: commits[0].root,
+                            path: path
+                        }
+                    ],
+                    excludeParents: true
+                };
+
+                return safeStorage.loadPaths(loadPathsParams);
+            })
+            .then(function (dataObjects) {
+                var hashes = Object.keys(dataObjects);
+                if (hashes.length === 1) {
+                    return dataObjects[hashes[0]];
+                } else if (hashes.length === 0) {
+                    throw new Error('Path does not exist ' + path);
+                } else {
+                    // This should never happen..
+                    throw new Error('safeStorage.loadPaths returned with more than one object');
+                }
+            });
+    }
+
     router.get('/projects', ensureAuthenticated, function (req, res, next) {
         var userId = getUserId(req);
         safeStorage.getProjects({username: userId})
@@ -539,7 +576,6 @@ function createAPI(app, mountPath, middlewareOpts) {
                 next(err);
             });
     });
-
 
     /**
      * Creating project by seed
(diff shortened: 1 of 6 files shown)


src/server/api/index.js
```diff
@@ -606,6 +642,48 @@ function createAPI(app, mountPath, middlewareOpts) {
...
+                    req.params.projectName),
+                commitHash = StorageUtil.getHashTaggedHash(req.params.commitHash);
+
+            loadNodePathByCommitHash(userId, projectId, commitHash, '/' + req.params[0])
+                .then(function (nodeObj) {
+                    res.json(nodeObj);
+                })
+                .catch(function (err) {
+                    if (err.message.indexOf('not exist') > -1 || err.message.indexOf('Not authorized to read') > -1 ) {
+                        err.status = 404;
```
kecso: not authorized status code is 403

src/server/storage/storage.js
```diff
@@ -357,35 +361,53 @@ Storage.prototype.loadPaths = function (data, callback) {
 
     this.mongo.openProject(data.projectId)
         .then(function (dbProject) {
-            var loadedObjects = {};
-
-            //self.logger.error('paths:', data.paths);
+            var loadedObjects = {},
+                throttleDeferred = Q.defer(),
```
kecso: what is saved here by having this 'throttle' function? shouldn't we use a library throttle (and for the loadObjects instead of the loadPath(s))?

test/server/storage/safestorage.spec.js
```diff
@@ -456,6 +458,22 @@ describe('SafeStorage', function () {
                 .catch(done);
         });
 
+        it('should getCommits and get the one specified', function (done) {
```
kecso: 'and' feels extra in the description

src/server/api/index.js
pmeijer: It was requested that if the user doesn't have read access the return code should be 404.
For two reasons;
The storage just checks if the user has the project in his/her project list (with read access) and does't check whether the project exists or not.
If the user doesn't have read access, he shouldn't even be able to get information about whether the project exists or not.

src/server/storage/storage.js
pmeijer: An example:
The root is loaded and the meta nodes are all requested to be loaded:
/languageNode/metaNodePath1
/languageNode/metaNodePath2
/languageNode/metaNodePathN

Without the throttle the languageNode will be loaded N-times from the mongo database. With the throttle it is only loaded once. The loadObject-calls are already throttled for a one path (or what do you mean?).

Q doesn't have a throttle and I wouldn't like to use another library.

test/server/storage/safestorage.spec.js
pmeijer: It is intended to be extra, it should not only get the commits, but also get the one specified.
But yes I agree, the phrasing is somewhat awkward. However since it was a getCommits test and the other such tests were named that way, I kept the 'should getCommits'.

src/server/api/index.js
kecso: thanks

test/server/storage/safestorage.spec.js
kecso: I see, okay then

src/server/storage/storage.js
kecso: I see. I thought you were trying to throttle against too many paralel calls towards the db, but it okay then.

PR #665, opened 2015-11-11
Property Editor accepts objects drops for pointers.

pmeijer:
The functionality is disabled for base pointers.


pmeijer:
Don't merge this yet

PR #666, opened 2015-11-11
Relative url methods in blob exec clients

pmeijer:
Fixes a browser bug for ExecutorPlugin.

Evil PR# :)


pmeijer:
Please wait with merge till I've squashed..

PR #672, opened 2015-11-13
Added support for click-and-drag. Fixes #633

### LINKED ISSUE #633: Moving objects
I cannot move objects without selecting them first. Is this by design? I think we were able to do it before and I find it very handy.

brollb:


src/client/js/Widgets/DiagramDesigner/SelectionManager.js
```diff
@@ -50,7 +50,6 @@ define([
         //enable SelectionManager specific DOM event listeners
         var self = this;
 
-        // note this is called only when the mouse was down and up
```
brollb: I removed this comment because it is no longer true.

Now it depends if the click is creating a new selection. If it is creating a new selection, it processes the click on mouse down. Otherwise, it processes it on mouse up (like normal). This allows the user to click-and-drag as desired.

PR #680, opened 2015-11-17
Fix corrupt diff generation

kecso:
During difference generation it can happen that the relation information regarding a node is only partial at the time of checking, but it should not cause exception.
The logging during merge was also improved so the server logs would now point to actual place of error.

### DIFF
--- a/src/common/core/corediff.js
+++ b/src/common/core/corediff.js
@@ -398,6 +398,8 @@ define(['common/util/canon',
                 tDiff = getPathOfDiff(diff, paths[i]);
                 if (tDiff.removed !== true) {
                     tDiff.pointer = tDiff.pointer || {source: {}, target: {}};
+                    tDiff.pointer.source = tDiff.pointer.source || {};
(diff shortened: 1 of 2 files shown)


pmeijer:
Any possibility to reproduce this and make a test?

kecso:
I was able to locate the difference among the two state of the project that resulted in failing diff generation. The change was a 'simple' base modification, but so far I was not able to reproduce the issue with another model that has the same 'change'.
So I verified the modification by checking the differences and the resulting merge, and they were correct.

pmeijer:
What does this mean: "...that the relation information regarding a node is only partial at the time of checking.."
Is it still possible that this info is only partial? Should it be partial?

pmeijer:
I assume this issue is related to the merge problem for the Genomics project. What will be the behavior after this change?

kecso:
It could happen that the empty portion is 'filtered' out before we try to extend our information based on the overlays.
The expected behavior is a successful mere of the branch (merge_from_here) into the branch (merge_to_here).

PR #682, opened 2015-11-18
Closes #676 RunOnServer toggle box not shown if readOnly

### LINKED ISSUE #676: Run plugin toggle should be removed if server execution is disabled
If the server side plugin execution is disabled, the toggle for server side execution (in the run plugin dialog) should be removed rather than disabled.

pmeijer:


brollb:
I think it would look better if it was italicized and/or greyed out. Thoughts?

PR #684, opened 2015-11-18
Refactored crosscut functionality.

kecso:
- connections now visualized like in Composition view
- set membership relation is visible
- relations can be filtered on every crosscut
- filter states are saved with the project
- default connection manager is set to CM2
- manual tests have been created and stored on google drive
- if a node is visualized as port and as a separate item as well the preference is at the item (so connections will end there)
- adding a node to the crosscut if it was already visualized as port is now possible (both become be visible)
- removing a node that is also visualized as port is now possible

### DIFF
--- a/src/client/js/Panels/ControllerBase/DiagramDesignerWidgetMultiTabMemberListControllerBase.js
+++ b/src/client/js/Panels/ControllerBase/DiagramDesignerWidgetMultiTabMemberListControllerBase.js
@@ -689,6 +689,7 @@ define(['js/logger',
 
         this._selectedMemberListMembers = [];
         this._delayedConnections = [];
+        this._delayedConnectionsAsItems = {};
 
         //remove current territory patterns
         if (this._selectedMemberListMembersTerritoryId) {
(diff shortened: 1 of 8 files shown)


src/client/js/Panels/ControllerBase/DiagramDesignerWidgetMultiTabMemberListControllerBase.js
```diff
@@ -1808,6 +1868,7 @@ define(['js/logger',
         this._GMEID2ComponentID[gmeID].push(uiComponent.id);
         this._ComponentID2GMEID[uiComponent.id] = gmeID;
 
+        this._delayedConnectionsAsItems = this._delayedConnectionsAsItems || {};
```
pmeijer: Couldn't this property be defined in the constructor? Is it declared here too? Or is it set to null at some point?

kecso: you are right, it should be in the constructor (or actually in the initialization of the tab...)

pmeijer:
See #668 regarding using CM2..

kecso:
I'll check the tests again with AR and remove the change of default CR if possible.

pmeijer:
Ok, is it ready to go?

kecso:
yepp, it should be good to go

PR #685, opened 2015-11-19
Closes #677 add more project info

### LINKED ISSUE #677: More information about the projects
It would be great if we could have more information about the projects in the projects' list. For instance, we can have the last modified time of the projects.

pmeijer:


src/client/js/Dialogs/Projects/ProjectsDialog.js
```diff
@@ -224,58 +222,114 @@ define([
...
+
+                if (result === 0) {
+                    if (rowAData._id.toUpperCase() > rowBData._id.toUpperCase()) {
+                        result = 1;
+                    } else {
+                        result = -1;
+                    }
+                }
+
+                return reverse ? result * (- 1) : result;
```
lattmann: smart!

PR #691, opened 2015-11-24
Fix issues in ProjectDialog and revise project creation.

pmeijer:
- Fix memory leak after ProjectDialog is closed.
- Fix safari bug when clicking buttons in ProjectDialog.
- Remove unintended eastern egg when project named "Something" in Projects dialog.
- Add gme icon in project dialog, add sort buttons and fix some style.
- Add a common dialog for creating project (using tabs for the different types).
- Fix the ProjectNavigator when clicking create/import - it should expand the the text box and put it in focus.
- Import project should upload the file to the blob and the seeding should take place in a worker.


pmeijer:
This should go both into master and v1.2.x

src/client/js/Dialogs/Projects/ProjectsDialog.js
```diff
@@ -648,33 +620,13 @@ define([
...
-    };
-
-    ProjectsDialog.prototype._createProjectFromSeed = function (projectName, type, seedName, branchName, commitHash) {
+    ProjectsDialog.prototype._createProjectFromSeed = function (
+        projectName,
+        type,
+        seedName,
+        branchName,
+        commitHash
+    ) {
```
kecso: is it intentional not to use the 'blobHash' parameter (it was set at the place of function call)?

pmeijer: I'm just reusing the seedName. It might not be the nicest way, however the seedName already changed meaning based on the 'db' and 'file' types (ProjectId vs file name).
With using it for a blob too, it should ideally been named seedId..

PR #692, opened 2015-11-24
Use promises for executor/blob/plugin/coreAPI

pmeijer:
Use coreQ everywhere. Add tests.

### DIFF
--- a/src/plugin/coreplugins/ExecutorPlugin/ExecutorPlugin.js
+++ b/src/plugin/coreplugins/ExecutorPlugin/ExecutorPlugin.js
@@ -15,8 +15,9 @@ define([
     'plugin/PluginBase',
     'executor/ExecutorClient',
     'common/util/ejs',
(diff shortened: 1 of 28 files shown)


src/plugin/coreplugins/ExecutorPlugin/ExecutorPlugin.js
```diff
@@ -211,32 +195,27 @@ define([
...
-            self.blobClient.getObject(newNameJsonHash, function (err, newName) {
+                return deferred.promise;
+            })
+            .then(function (jobInfo) {
+                self.logger.info('job succeeded', jobInfo);
+                finalJobInfo = jobInfo;
+                return self.blobClient.getMetadata(jobInfo.resultHashes.resultFile);
+            })
+            .then(function (metaData) {
+                console.log(metaData);
```
kecso: is this log necessary?

PR #694, opened 2015-11-25
Fixes the issue of parsing large character buffer into string

kecso:
as the library function can cause stack overflow


kecso:
please cherry-pick the correction into v1.2.x as well

PR #695, opened 2015-11-25
Fixes stack overflow exception during project import

kecso:
The library function of character array conversion to string can cause stack overflow so we implemented our own.
Now every user of String.fromCharCode who intends to use it with larger array, should use the own utility function uin8ArrayToString

### DIFF
--- a/src/common/blob/BlobClient.js
+++ b/src/common/blob/BlobClient.js
@@ -8,413 +8,413 @@
  * @author ksmyth / https://github.com/ksmyth
  */
 
-define(['blob/Artifact',
-        'blob/BlobMetadata',
-        'superagent',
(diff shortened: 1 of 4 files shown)


kecso:
only the first commit of the branch should be cherry-picked to v1.2.x branch as the other commit contains modification that targets code that will be available only in 1.3.0

pmeijer:
Can you fix the BlobClient diff? You added an extra indentation...
Do like this:
```
define([
    'common/storage/storageclasses/editorstorage',
    'common/storage/socketio/browserclient',
    'common/storage/socketio/websocket',
], function (EditorStorage, BrowserIoClient, WebSocket) {
    'use strict';
```

PR #699, opened 2015-11-26
Fixed ProjectDialog exception after delete.

pmeijer:
Remove css source-maps.


pmeijer:
Merge into master and cherry pick to 1.2.1.

PR #705, opened 2015-11-30
Fixes issue, that 'Show all...' on navigator bar causes exception

kecso:


kecso:
The change should be cherry-picked to v1.2.x branch.

PR #709, opened 2015-11-30
Plugin can be invoked on read-only and commit.

pmeijer:
This branch also introduces a new branch status ERROR.

When running a plugin in the client the storage is shared between the plugin and client. When a plugin attempts to make a change into a branch on a read-only project - the client updates its state to include the changes before the server responds with an error. These errors (makeCommit) are now propagated to the users via the ERROR branch-status, which prompts the user to re-select the branch.

### DIFF
--- a/src/client/js/Dialogs/PluginResults/PluginResultsDialog.js
+++ b/src/client/js/Dialogs/PluginResults/PluginResultsDialog.js
@@ -122,6 +122,7 @@ define(['js/util',
                 } else if (commits[c].status === storageUtil.CONSTANTS.SYNCED) {
                     message = 'updated';
                 } else if (!commits[c].status) {
+                    // When null or undefined there was no branch target for the commit..
                     message = 'made commit';
                 } else {
                     message = 'forked and created';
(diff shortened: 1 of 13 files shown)


src/plugin/PluginManagerBase.js
```diff
@@ -249,11 +249,11 @@ define(['plugin/PluginBase', 'plugin/PluginContext', 'common/storage/util'],
                     return;
                 }
                 self.logger.debug(branches);
-                if (branches.hasOwnProperty(managerConfiguration.branchName)) {
+                if (managerConfiguration.branchName && !branches.hasOwnProperty(managerConfiguration.branchName)) {
```
lattmann: use === false instead of !

src/common/storage/project/branch.js
```diff
@@ -182,7 +182,14 @@ define(['common/storage/constants'], function (CONSTANTS) {
             var i;
 
             logger.debug('dispatchBranchStatus old, new', branchStatus, newStatus);
-            branchStatus = newStatus;
+
+            if (branchStatus === CONSTANTS.BRANCH_STATUS.ERROR) {
+                logger.warn('In error state, action from user required');
```
lattmann: log this as an error, maybe

src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
```diff
@@ -128,6 +129,12 @@ define([
                         });
                     }
                 });
+            } else if (value === ITEM_VALUE_SELECT_BRANCH) {
+                self._client.selectBranch(branchName, null, function (err) {
+                    if (err) {
+                        self._logger.error(err);
+                    }
+                });
```
lattmann: add `self._logger.debug('branch selected: ' + branchName);`

src/client/js/Dialogs/PluginResults/PluginResultsDialog.js
```diff
@@ -121,6 +121,8 @@ define(['js/util',
                     message = 'started from';
                 } else if (commits[c].status === storageUtil.CONSTANTS.SYNCED) {
                     message = 'updated';
+                } else if (!commits[c].status) {
```
lattmann: add a comment:
`// if status is falsy`

PR #729, opened 2015-12-09
Added autorouter replay download button. Fixes #728

### LINKED ISSUE #728: AutoRouter Replay Download Button
I think it would be useful if, while in debug mode, the DiagramDesigner provided a download button for the autorouter replay file in the toolbar.

Thoughts?

brollb:
Let me know if you think it should be added at the end behind a separator or something. Currently, the download button is right after the search field

### DIFF
--- a/src/client/js/Widgets/DiagramDesigner/DiagramDesignerWidget.js
+++ b/src/client/js/Widgets/DiagramDesigner/DiagramDesignerWidget.js
@@ -206,15 +206,12 @@ define([
 
         /*********** CONNECTION DRAWING COMPONENT *************/
         this._defaultConnectionRouteManagerType = params.defaultConnectionRouteManagerType;
-        //initiate Connection Router (if needed)
+
+        //initiate Connection Router
         if (params.connectionRouteManager) {
             this.connectionRouteManager = params.connectionRouteManager;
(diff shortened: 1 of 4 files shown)


src/client/js/Widgets/DiagramDesigner/DiagramDesignerWidget.js
```diff
@@ -206,16 +207,10 @@ define([
...
-            this.connectionRouteManager = new ConnectionRouteManagerBasic({diagramDesigner: this});
-        } else if (params.defaultConnectionRouteManagerType === 'basic2') {
-            this.connectionRouteManager = new ConnectionRouteManager2({diagramDesigner: this});
-        } else if (params.defaultConnectionRouteManagerType === 'basic3') {
-            this.connectionRouteManager = new ConnectionRouteManager3({diagramDesigner: this});
-        }
+
+        //initiate Connection Router
+        router = params.connectionRouteManager || params.defaultConnectionRouteManagerType;
+        this._onConnectionRouteManagerChanged(router);
```
pmeijer: Wouldn't this change the behavior if params.connectionRouteManager is passed?
That's an instance of a RouteManager and in _onConnectionRouteManagerChanged the switch statement would always reset it to ConnectionRouteManagerBasic.

PR #732, opened 2015-12-09
Added animation functionality. Fixes #725

### LINKED ISSUE #725: Smooth Connection Updates
It would be cool if the autorouter provided d3 transitions for the connection path updates

brollb:


brollb:
I need to fix the location of the connection name. Don't merge this yet.

However, feedback on this feature would be appreciated

brollb:
Fixed the location of the names

pmeijer:
Brian I'm a little bit uncertain about the timeouts... Are they related to the issue we saw before?

pmeijer:
I will merge it and we'll see what people think and how stable it is. I really like the look of it, but as we talked about, if it jeopardizes stability and speed we might have to revoke it.

PR #741, opened 2015-12-10
Add method isInvalidActiveNode on PluginBase.

pmeijer:
This adds a convenience method to check if the plugin is registered for the activeNode.
Although it could be checked by default from the pluginManagers, there are cases where this would be cumbersome. For instance most of our tests would have to be revisited.


lattmann:
Naming of the API function may not be the best, but I can see why you have chosen this name.

PR #755, opened 2015-12-17
Fixes #749 Use the common regular expressions throughout the UI.

### LINKED ISSUE #749: Use the common regular expressions throughout the UI.
Refine if needed.
Add regexp for e.g. attributes in the meta-editor.

kecso:
- now every key (attribute, constraint, aspect) on the META visualizer
- if the name is not valid the text becomes red
- in the project history branches and tags can have number as name


pmeijer:
Please make the text-box border red instead of the text when there is an error.

PR #778, opened 2016-01-07
Fixes #772 Instance target bug

### LINKED ISSUE #772: Instance target bug
When I'm importing a project, I'm getting tens of ```error handling needed???!!!???``` messages. Eventually, it looks like it imports the project on the client side. However, the imported project has broken connections. You can check ROOT > apps > Blink > BlinkAppC for an example after importing this: https://drive.google.com/file/d/0B6icounl9q69akgwdnEzaFFHLTA/view?usp=sharing

kecso:
The commits should be cherry-picked into the master as well.

### DIFF
--- a/src/common/core/users/serialization.js
+++ b/src/common/core/users/serialization.js
@@ -188,7 +188,7 @@ define(['common/util/assert', 'blob/BlobConfig'], function (ASSERT, BlobConfig) 
                 };
 
             for (i = 0; i < keys.length; i += 1) {
-                if (typeof  object[keys[i]] === 'string' && object[keys[i]].indexOf('/') === 0) {
+                if (typeof object[keys[i]] === 'string' && object[keys[i]].indexOf('/') === 0) {
                     object[keys[i]] = pathToCompositeId(object[keys[i]]);
                 } else if (typeof object[keys[i]] === 'object' && object[keys[i]] !== null) {
                     pathToGuidInObject(object[keys[i]]);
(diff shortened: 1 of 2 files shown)


src/common/core/users/serialization.js
```diff
@@ -1031,6 +1061,8 @@ define(['common/util/assert', 'blob/BlobConfig'], function (ASSERT, BlobConfig)
                     core.setPointer(node, keys[i], null);
                 } else if (nodes[target] && toRemoveGuids.indexOf(target) === -1) {
                     core.setPointer(node, keys[i], nodes[target]);
+                } else if (isCombinedId(target)) {
+                    core.setPointer(node, keys[i], getCombinedTarget(target));
                 } else {
                     console.log('error handling needed???!!!???');
```
pmeijer: How about bringing in that logger and/or throwing an exception here?

src/common/core/users/serialization.js
```diff
@@ -171,13 +170,26 @@ define(['common/util/assert', 'blob/BlobConfig'], function (ASSERT, BlobConfig)
...
+
+                    return relPath === '' ? guid : guid + '@' + relPath;
+                };
 
             for (i = 0; i < keys.length; i += 1) {
-                if (typeof object[keys[i]] === 'string' &&
-                    object[keys[i]].indexOf('/') === 0 &&
-                    (ancestorPathToGuid[object[keys[i]]] || pathToGuid[object[keys[i]]])) {
-                    object[keys[i]] = ancestorPathToGuid[object[keys[i]]] || pathToGuid[object[keys[i]]];
+                if (typeof  object[keys[i]] === 'string' && object[keys[i]].indexOf('/') === 0) {
```
pmeijer: additional space...

pmeijer:
Looks good besides two minor notes. Coverage looks good too.

src/common/core/users/serialization.js
kecso: I was checking the logger, but either the public API should be changed, or some global logger should be created.
On the other hand the given else branch is considered (again) a never run into branch, that is why I left it that way.

pmeijer: How about throwing an exception?
It was probably meant not to be entered before either but it did.
Now if someone imports and doesn't pay attention to the logs they might end up with an incomplete model

hakantunc:
Thanks a bunch

PR #780, opened 2016-01-08
Add plugin for developers to evaluate and debug constraints.

pmeijer:
Plugin for facilitating construction of constraints. This plugin is mainly intended for developers.
There are three different modes to execute the plugin:

#### EvaluateConstraints
Loads the constraints defined in "Constraints.js" and applies them to all nodes starting from the activeNode. If the constraint returns an error or an exception is thrown it is reported as an error and a message with the full stack is added to the result. If the constraint succeeds or fails, it is simply reported in the summary.

#### GenerateConstraints
Goes through all the meta-nodes in the model and extracts the constraints and generates a  "Constraints.js". This is intended to be used to overwrite the file and run/tweak and write the fixed constraints back to the model.

#### PopulateFromConstraints
Reads in the constraints from "Constraints.js" and adds them to the matched meta-nodes in the model. N.B. if the constraint already exists it will be overwritten by the new one.


src/client/decorators/MetaDecorator/DiagramDesigner/styles/ConstraintDetailsDialog.css
```diff
@@ -3,3 +3,13 @@
...
+  .constraint-details-dialog .modal-body label.control-label {
+    width: 15%; }
+  .constraint-details-dialog .modal-body div.controls {
+    width: 80%; }
+    .constraint-details-dialog .modal-body div.controls input.name-input {
+      width: 30%; }
+    .constraint-details-dialog .modal-body div.controls input.info-input {
+      width: 60%; }
+
+/*# sourceMappingURL=ConstraintDetailsDialog.css.map */
```
pmeijer: I will fix this

PR #782, opened 2016-01-11
Executor uses mongodb and worker can be queried for output. Closes #686

### LINKED ISSUE #686: Executor feedback during execution
It would be useful if executors could be queried while they are running for their stdout and stderr logs. This could provide the users progress feedback, etc.

pmeijer:


src/common/executor/ExecutorOutputQueue.js
```diff
@@ -0,0 +1,120 @@
...
+        /**
+         * Adds a string to the current output (which will be queued based on interval and/or segmentSize).
+         * @param {string} outputStr
+         */
+        this.addOutput = function (outputStr) {
+            var lines = outputStr.split(/\r\n|\r|\n/);
+
+            self.currOutput.lines = self.currOutput.lines.concat(lines);
+            self.logger.debug('length', self.currOutput.lines.length);
+            if (segmentSize > -1 && self.currOutput.lines.length >= segmentSize) {
```
kecso: shouldn't it be <=segmentSize? 
(and if it is, shouldn't we avoid 0 segmentSize?)

src/plugin/coreplugins/ExecutorPlugin/ExecutorPlugin.js
```diff
@@ -178,17 +179,42 @@ define([
...
+                                            var completeOutput = '';
+                                            output.forEach(function (o) {
+                                                completeOutput += o.output;
+                                            });
+                                            self.logger.debug('complete output\n', completeOutput);
+                                            deferred.resolve(jInfo);
+                                        })
+                                        .catch(deferred.reject);
+                                } else {
+                                    console.log('no output after success');
```
kecso: is it necessary to be console-log?

src/common/executor/ExecutorOutputQueue.js
pmeijer: Less or equal? Then it would always send output no matter how many lines were specified...

0 is avoided in the ExecutorWorker, see:

https://github.com/webgme/webgme/pull/782/files#diff-605c18a4cfb5b750e60ec51ec9737f42R214

and from the client:

https://github.com/webgme/webgme/pull/782/files#diff-3d5a46c06fec80e5ac136973df3d164cR78

kecso: yeah, I got it backwards.

PR #788, opened 2016-01-15
Fixes #785 URL query for node is not working for paths with non-numbers.

### LINKED ISSUE #785: URL query for node is not working for paths with non-numbers.

kecso:
Now a library is used for uri query string parsing.
The node parameter of the query is not transformed to lower case as it can have both upper and lower case characters now.


src/client/js/main.js
```diff
@@ -78,6 +78,8 @@ require.config({
 
         moment: 'lib/moment/moment.min',
 
+        urlparse:'lib/purl/purl.min',
```
lattmann: Add space next time.

src/client/js/util.js
```diff
@@ -105,12 +105,13 @@ define([], function () {
         },
 
         getURLParameterByName: function (name) {
-            var regex,
-                results;
-            name = name.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]');
-            regex = new RegExp('[\\?&]' + name + '=([^&#]*)');
-            results = regex.exec(location.search);
-            return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
+            var queryParams = URLPARSE(location,true).param();
```
lattmann: Style, space is missing between function parameters.

PR #789, opened 2016-01-15
Patch Root Communication feature

kecso:
if config.storage.patchRootCommunicationEnabled is set to true the makeCommit will contain only a jsonPatch type root
in every caches and storages the complete root object is saved though
only makeCommit and BRANCH_UPDATED event can contain the jsonPatch type root


src/common/storage/storageclasses/editorstorage.js
```diff
@@ -223,22 +225,67 @@ define([
 
                 branch._remoteUpdateHandler = function (_ws, updateData, initCallback) {
                     var j,
-                        originHash = updateData.commitObject[CONSTANTS.MONGO_ID];
+                        inserts = [],
+                        insertFailed = false,
+                        originHash = updateData.commitObject[CONSTANTS.MONGO_ID],
+                        insertObject = function (coreObject) {
```
pmeijer: Make a separate base class, "PatchStorage" for all this business.

src/common/storage/storageclasses/editorstorage.js
```diff
@@ -223,22 +225,67 @@ define([
...
+                                    project.insertObject(coreObject);
+                                    deferred.resolve();
+                                }
+                            }
+
+                            return deferred.promise;
+                        };
                     logger.debug('_remoteUpdateHandler invoked for project, branch', projectId, branchName);
                     for (j = 0; j < updateData.coreObjects.length; j += 1) {
-                        project.insertObject(updateData.coreObjects[j]);
```
pmeijer: self.insertObjects()
   .then(...)
   .catch( emit branch error)

src/common/storage/storageclasses/editorstorage.js
```diff
@@ -351,41 +398,65 @@ define([
                     projectId: projectId,
                     commitObject: null,
                     coreObjects: null
+                },
+                rootHash = Array.isArray(rootHash_es) ? rootHash_es[0] : rootHash_es,
+                baseRootHash = Array.isArray(rootHash_es) ? rootHash_es[1] : null,
+                fillPatchRoot = function (next) {
```
pmeijer: Move this to that class too

src/common/storage/storageclasses/editorstorage.js
```diff
@@ -351,41 +398,65 @@ define([
...
+                    if (gmeConfig.storage.patchRootCommunicationEnabled &&
+                        baseRootHash !== null &&
+                        coreObjects[rootHash] &&
+                        project) {
+                        project.loadObject(baseRootHash, function (err, baseRoot) {
+                            if (!err && baseRoot) {
```
pmeijer: No Error handling??
Emit branch Error!

src/server/storage/storage.js
```diff
@@ -165,12 +166,39 @@ Storage.prototype.makeCommit = function (data, callback) {
...
-                rootProvided;
+                rootProvided,
+                patchRoot = null;
 
             function insertObj(hash) {
                 return project.insertObject(data.coreObjects[hash]);
             }
 
+            function handlePatchRoot() {
+                var pathcRootDeferred = Q.defer();
```
pmeijer: variable name..

src/common/storage/storageclasses/patchstorage.js
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    function _insertObject(project, coreObject) {
+        var deferred = Q.defer(),
+            patchObject;
+        if (coreObject) {
+            if (coreObject.type === 'patch') {
+                patchObject = coreObject;
+                project.loadObject(patchObject.base, function (err, baseObject) {
+                    if (err || !baseObject) {
```
pmeijer: Thought it wasn't necessary to check the baseObject here

src/common/storage/storageclasses/editorstorage.js
```diff
@@ -351,41 +354,46 @@ define([
...
-            if (branch) {
-                logger.debug('makeCommit, branch is open will commit using commitQueue. branchName:', branchName);
-                self._commitToBranch(projectId, branchName, commitData, parents[0], commitCallback);
-            } else {
-                webSocket.makeCommit(commitData, commitCallback);
-            }
+                    logger.debug('makeCommit', commitData);
+                    if (branch) {
+                        logger.debug('makeCommit, branch is open will commit using commitQueue. branchName:', branchName);
+                        self._commitToBranch(projectId, branchName, commitData, parents[0], commitCallback);
```
pmeijer: The introduced asynchronousy will be an issue with multiple changes

src/common/storage/project/cache.js
```diff
@@ -23,12 +27,28 @@ define(['common/util/assert', 'common/storage/constants'], function (ASSERT, CON
...
         function cacheInsert(key, obj) {
+            var patch;
             ASSERT(obj[CONSTANTS.MONGO_ID] === key);
             logger.debug('cacheInsert', key);
 
             //deepFreeze(obj);
             if (!cache[key]) {
-                cache[key] = obj;
+                if(obj.type === 'patch'){
+                    if(cache[obj.base] || backup[obj.base]){
```
pmeijer: Would it make sense to call `getFromCache` here?
Or the only time a patch object is inserted is when we receive it from another user (so it couldn't be in the makeCommit queue..)?

package.json
```diff
@@ -60,6 +60,7 @@
     "socket.io": "=1.3.2",
     "socket.io-client": "=1.3.2",
     "superagent": "=1.4.0",
+    "underscore": "^1.8.3",
```
pmeijer: Tie down version

src/common/storage/storageclasses/editorstorage.js
```diff
@@ -40,7 +40,7 @@ define([
         self.logger = logger;
         self.userId = null;
 
-        StorageObjectLoaders.call(this, webSocket, mainLogger, gmeConfig);
+        PatchStorage.call(this, webSocket, mainLogger, gmeConfig);
 
         this.open = function (networkHandler) {
```
pmeijer: What happened with the insert of the received patches? Am I missing something?

src/common/storage/socketio/websocket.js
```diff
@@ -174,6 +174,19 @@ define([
...
+
+            if (gmeConfig.storage.patchRootCommunicationEnabled && data.patchRoot) {
+                data.coreObjects = {};
+                for (i = 0; i < keys.length; i += 1) {
+                    data.coreObjects[keys[i]] = localCoreObjects[keys[i]];
+                }
+                data.coreObjects[data.commitObject.root] = data.patchRoot;
+            }
+            delete data.patchRoot; //we remove the field in either way
+
```
pmeijer: I'm not sure about his part..
The commitQueue has a reference to `data`, which has a reference to `data.coreObjects`.
Even though `localCoreObjects` still contains the full root object, there is no reference to it...
So if the cache is cleared, the rootNode is not available anywhere.

kecso: Ohh, you are right. I tought we only have a reference to the coreObjects in the queue.
I will copy/rebuild the whole data then and send that out, so we will still have the reference in the queue

pmeijer: I was mistaken.. In the cache we don't reference the `commitQueue`/`data`, but rather the `coreObjects` directly.
So never mind my comment

pmeijer: However, we need to keep this in mind if we add functionality to dump the local changes after a disconnect (and no automatic reconnect).

kecso: Definitely.
However, if we dump every 'hanging' commit and 'replay' them on the server before reconnection, it should not be a problem.

src/client/js/client.js
```diff
@@ -151,6 +152,13 @@ define([
                         logger.warn('Lots of persisted objects', numberOfPersistedObjects);
                     }
 
+                    if (gmeConfig.storage.patchRootCommunicationEnabled) {
+                        persisted.rootHash = {
```
pmeijer: Couldn't these be provided from the core?
So when enabled, instead of returning just `{rootHash: '#..', objects: {}}`, it also adds rootInfo?

src/common/storage/storageclasses/patchstorage.js
```diff
@@ -0,0 +1,41 @@
...
+        this.logger = this.logger || logger.fork('patch');
+    }
+
+    PatchStorage.prototype = Object.create(StorageObjectLoaders.prototype);
+
+    PatchStorage.prototype.constructor = PatchStorage;
+
+    PatchStorage.prototype._fillPatchRoot = function (commitData, rootInfo) {
+
+        if (this.gmeConfig.storage.patchRootCommunicationEnabled &&
```
pmeijer: I think `this.gmeConfig.storage.patchRootCommunicationEnabled` should only be applied at two points.
In the core when returing rootHash/rootInfo and on the server side when dispatching the event.

src/client/js/client.js
pmeijer: Btw Plugins, AddOns, client/merge.js, core/users/merge.js are not updated to use the patch-root.

src/common/core/coretree.js
```diff
@@ -716,6 +717,10 @@ define([
                 result = {rootHash: node.data[ID_NAME], objects: {}};
             }
 
+            if (gmeConfig.storage.pathRootCommunicationEnabled && getRoot(node) === null) {
+                node.baseRawData = node.data;
```
pmeijer: Does this work well with the first commit?

src/common/storage/constants.js
```diff
@@ -48,6 +48,22 @@
...
  */
+
+/**
+ * @typedef {object} PatchObject
+ * @prop {module:Core~ObjectHash} _id - Hash of the expected result object.
+ * @prop {module:Core~ObjectHash} base - Hash of the base object where the patch should be applied.
+ * @prop {string} type - 'patch'.
+ * @prop {object} patch - The patch instructions (based on [RFC6902]{@link http://tools.ietf.org/html/rfc6902}).
+ *
+ * @example
```
pmeijer: +1

package.json
```diff
@@ -60,6 +60,7 @@
     "socket.io": "=1.3.2",
     "socket.io-client": "=1.3.2",
     "superagent": "=1.4.0",
+    "underscore": "=1.8.3",
```
pmeijer: Can't you harmonize the version with the one at the client side?

src/common/core/coretree.js
```diff
@@ -689,6 +690,9 @@ define([
 
             node = normalize(node);
 
+            //currently there is no reason to call the persist on a non-root object
+            node = getRoot(node);
+
```
pmeijer: Doesn't this mean that we always get the root, but it is not necessarily normalized?

pmeijer: Or normalize works upwards in the containment chain?

src/common/storage/project/cache.js
```diff
@@ -250,6 +254,32 @@ define(['common/util/assert', 'common/storage/constants'], function (ASSERT, CON
...
+                        patch.result[CONSTANTS.MONGO_ID] = key;
+                        this.insertObject(patch.result);
+                    } else {
+                        logger.error('patch application failed', patch);
+                    }
+                } else {
+                    logger.debug('the base is not available from the cache so the patch object is ignored');
+                }
+            } else {
+                logger.error('invalid patch object format');
```
pmeijer: log the obj too here.

src/common/storage/project/cache.js
```diff
@@ -250,6 +254,32 @@ define(['common/util/assert', 'common/storage/constants'], function (ASSERT, CON
...
+            if (obj.base && typeof obj.patch === 'object' && key) {
+                base = getFromCache(obj.base);
+
+                if (base) {
+                    patch = jsonPatcher.apply(base, obj.patch);
+                    if (patch.status === 'success') {
+                        patch.result[CONSTANTS.MONGO_ID] = key;
+                        this.insertObject(patch.result);
+                    } else {
+                        logger.error('patch application failed', patch);
```
pmeijer: log obj here too

src/common/storage/storageclasses/editorstorage.js
```diff
@@ -342,19 +347,39 @@ define([
             commitNext();
         };
 
-        this.makeCommit = function (projectId, branchName, parents, rootHash, coreObjects, msg, callback) {
+        this.makeCommit = function (projectId, branchName, parents, rootHash, dataObjects, msg, callback) {
```
pmeijer: I see why you don't want it to be called coreObjects. However now there's an inconsistency between project api and this.

src/common/core/coretree.js
kecso: yeah, normalize works updwards
so though we agreed that a persist should always go on the complete containment, I though it would be nice to still start the normalization from the input node

src/common/storage/project/cache.js
kecso: I added the patch object to the patch result so it can be easily 'used' in other places as well.

PR #790, opened 2016-01-15
Closes #787 Copying multiple objects doesn't work

### LINKED ISSUE #787: Copying multiple objects doesn't work
When you select multiple objects in model editor and try to copy them with ctrl+drag, new FCO instances show up instead of copies of the original objects.

pmeijer:


test-karma/client/js/client.spec.js
```diff
@@ -1569,28 +1569,37 @@ describe('GME client', function () {
...
+            var rootNode = client.getNode(''),
+                startChildren = rootNode.getChildrenIds().length,
+                endChildren,
+                params = {parentId: ''};
+
+            params[clientNode.getId()] = {};
+            client.copyMoreNodes(params);
+
+            endChildren = rootNode.getChildrenIds().length;
+            expect(endChildren).to.equal(startChildren + 1);
```
kecso: I would also expect that the copy have the same registry/attributes as the original node.

PR #791, opened 2016-01-15
reassing_relids tool have been implemented

kecso:
with the refactored relid generation it is worth to reallocate relids to existing nodes in an export format so that the new smaller size can be utilized
the tool had been manually tested (as it is not intended for long-term usage).


src/bin/reassign_relids.js
```diff
@@ -0,0 +1,162 @@
+/* jshint node:true */
+/**
```
pmeijer: Add some description

src/bin/reassign_relids.js
```diff
@@ -0,0 +1,162 @@
...
+    program
+        .version('0.1.0')
+        .usage('<project-file> [options]')
+        .option('-o, --output [string]', 'the output file [by default, the input is overwritten]')
+        .parse(argv);
+
+    outputPath = program.output || program.args[0];
+
+    Q.nfcall(function () {
+            console.log('loading input');
```
pmeijer: logger.info

src/bin/reassign_relids.js
```diff
@@ -0,0 +1,162 @@
...
+                    for (i = 0; i < keys.length; i += 1) {
+                        if (typeof jsonNode[keys[i]] === 'string') {
+                            jsonNode[keys[i]] = getConvertedStringField(jsonNode[keys[i]]);
+                        } else if (typeof jsonNode[keys[i]] === 'object' && jsonNode[keys[i]] !== null) {
+                            nodeCompositeIdUpdate(jsonNode[keys[i]]);
+                        }
+                    }
+                };
+
+            console.log('converting input');
```
pmeijer: logger

PR #799, opened 2016-01-20
Fixes core node children cache handling

kecso:
- when a children is deleted from a node its entry in the children cache is emptied only allowing it to be reused, but during child creation, the cache was not checked for such objects
- when the client copies multiple object it should preserve the relative ids and paths of the source nodes
- all these errors were revealed by the new relative id generation, as it increased the probability of collision (which should be handled in all cases)

the fix should be cherry-picked into the master as well

### DIFF
--- a/src/common/core/coretree.js
+++ b/src/common/core/coretree.js
@@ -337,20 +337,7 @@ define([
                 throw new Error('invalid node data');
             }
 
-            var relid = RANDOM.generateRelid(node.data);
-            var child = {
-                parent: node,
-                relid: relid,
-                age: 0,
-                children: [],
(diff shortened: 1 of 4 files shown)


src/common/core/coretree.js
```diff
@@ -337,20 +337,7 @@ define([
...
-                children: [],
-                data: __getEmptyData()
-            };
-
-            // TODO: make sure that it is not on the list
-            node.children.push(child);
-
-            __ageRoots();
-            return child;
+            return getChild(node,RANDOM.generateRelid(node.data));
```
pmeijer: style..

PR #800, opened 2016-01-20
Optimized import project functionality.

kecso:
As the import use the same underlying function as the library update, it always cleared out the sets and meta rules and pointers before inserting the new ones. Due to the nature of the deletion process it takes some time even if there is nothing to remove, so to improve the overall speed, we check if the given node is a new one, and in that case we skip the remove.
The bigger a project is the bigger the positive impact of the change (example project was of size 8.7Mb, the import time decreased from 45s to 10s).
It also improves library updates where the number of new items is big.

### DIFF
--- a/test-karma/client/js/client.spec.js
+++ b/test-karma/client/js/client.spec.js
@@ -4313,12 +4313,14 @@ describe('GME client', function () {
                         superagent.get(url, function (err, result) {
                             expect(err).to.equal(null);
 
-                            expect(result.body).to.deep.equal(refNodeProj);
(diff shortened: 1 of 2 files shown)


test-karma/client/js/client.spec.js
```diff
@@ -4313,12 +4313,14 @@ describe('GME client', function () {
                         superagent.get(url, function (err, result) {
                             expect(err).to.equal(null);
 
-                            expect(result.body).to.deep.equal(refNodeProj);
```
pmeijer: Out of curiosity, does this this change really matter?

PR #804, opened 2016-01-26
Fixes #802 Multiple Meta views in Split View changes the tabs unexpectedly.

### LINKED ISSUE #802: Multiple Meta views in Split View changes the tabs unexpectedly.
Any changes (moving or adding any feature) made while two different meta sheets are open with a split view causes one of the meta sheets to load the other sheet and they show the same meta sheet.

kecso:
In split view, both half was updated according the global state value, however only the active one should follow the state.

### DIFF
--- a/src/client/js/Panels/MetaEditor/MetaEditorControl.js
+++ b/src/client/js/Panels/MetaEditor/MetaEditorControl.js
@@ -1762,8 +1762,8 @@ define(['js/logger',
             if (WebGMEGlobal.PanelManager._activePanel === this) {
                 selectedSheetID = WebGMEGlobal.State.get(CONSTANTS.STATE_ACTIVE_TAB);
             } else {
-                for(selectedSheetID in this._sheets || {}){
-                    if(this._sheets[selectedSheetID] === this._selectedMetaAspectSet){
+                for (selectedSheetID in this._sheets || {}) {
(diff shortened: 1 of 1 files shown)


src/client/js/Panels/MetaEditor/MetaEditorControl.js
```diff
@@ -1783,7 +1758,16 @@ define(['js/logger',
         if (WebGMEGlobal.State.get(CONSTANTS.STATE_ACTIVE_TAB) !== null &&
             WebGMEGlobal.State.get(CONSTANTS.STATE_ACTIVE_TAB) !== undefined &&
             metaAspectSheetsRegistry.length > WebGMEGlobal.State.get(CONSTANTS.STATE_ACTIVE_TAB)) {
-            selectedSheetID = WebGMEGlobal.State.get(CONSTANTS.STATE_ACTIVE_TAB);
+            //only the active panel should react to the global state
+            if (WebGMEGlobal.PanelManager._activePanel === this) {
+                selectedSheetID = WebGMEGlobal.State.get(CONSTANTS.STATE_ACTIVE_TAB);
+            } else {
+                for(selectedSheetID in this._sheets || {}){
```
pmeijer: Style...

Looks good otherwise (thought you explicitly were parsing the URL when we talked in the kitchen).

PR #805, opened 2016-01-26
Fix test cases with non-deterministic results.

kecso:
- copy of a node do not preserve the relid
- random relid generation should not be tested (only to prove that it always generates a relid)


lattmann:
I am not sure if this is the right fix or not. It is functional though.

PR #813, opened 2016-02-01
Core layers consistency

pmeijer:
First go on making a consistent style on the core layers.

The inner core passed should be named innerCore, the passed on core should be named core.

Private helper functions should be function declarations at the top (Helper Functions).

Overwritten functions should be "Modified Methods" and new/added functions should be "Added Methods". This distinction is partially done (there is no longer any intermediate local variables at least).

### DIFF
--- a/src/common/core/constants.js
+++ b/src/common/core/constants.js
@@ -12,6 +12,35 @@ define([], function () {
         OVERLAYS_PROPERTY: 'ovr',
         COLLECTION_NAME_SUFFIX: '-inv',
         ALL_SETS_PROPERTY: '_sets',
-        MEMBER_RELATION: 'member'
+        SET_MODIFIED_REGISTRY: '_sets_',
+        MEMBER_RELATION: 'member',
+        BASE_POINTER: 'base',
+
+        NULLPTR_NAME: '_null_pointer',
+        NULLPTR_RELID: '_nullptr',
+
+        META_SET_NAME: 'MetaAspectSet',
+        NULL_GUID: '00000000-0000-0000-0000-000000000000',
+        OWN_GUID: '_relguid',
+
+        CONSTRAINTS_RELID: '_constraints',
+        C_DEF_PRIORITY: 1,
+        CONSTRAINT_REGISTRY_PREFIX: '_ch#_',
+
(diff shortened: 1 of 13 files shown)


src/common/core/constants.js
```diff
@@ -12,6 +12,35 @@ define([], function () {
...
+
+        SET_ITEMS: 'items',
+        SET_ITEMS_MAX: 'max',
+        SET_ITEMS_MIN: 'min',
+
+        META_ASPECTS: 'aspects',
+        META_CHILDREN: 'children',
+        META_NODE: '_meta',
+        META_POINTER_PREFIX: '_p_',
+        META_ATTRIBUTE_PREFIX: '_a_',
```
kecso: sorry I was wrong, this prefix is used for ASPECTS

src/common/core/coretree.js
```diff
@@ -24,23 +24,17 @@ define([
...
-
+            logger = options.logger.fork('core'),
             ID_NAME = storage.ID_NAME,
-            __getEmptyData = function () {
-                return {};
-            },
             roots = [],
             ticks = 0,
+            mutateCount = 0,
+            checkValidTreeRunning = true,
```
kecso: this variable seems to be out of use, we should probably remove it

src/common/core/metacachecore.js
```diff
@@ -132,28 +153,22 @@ define(['common/util/assert', 'common/core/tasync'], function (ASSERT, TASYNC) {
...
-                }
-            }
-
-            //parameters
-            // node - the node in question
-            // children - the current children of the node, so that multiplicity can be checked
-            // sensitive - if true the function do not return the connection and abstract types
-            // multiplicity - if true the function filters out possibilities that fail multiplicity check
-            // aspect - if given the function also filters out valid children type meta nodes based on aspect rule
+            /**
```
kecso: this comment can be removed as the function description is done in the core.js

src/common/core/constants.js
pmeijer: META_ASPECT_PREFIX then?

kecso: yes, that would be perfect

PR #815, opened 2016-02-02
TreeBrowser added filters, editing, sorting and locate.

pmeijer:
1) Edit nodes is back again. Right-click or f2.
2) Sorting puts connections at bottom.
3) Filter form added. Show/Hide connections, abstracts and/or leaf nodes. Search filter based on name and meta-type.
4) Locate button in toolbar. (Requires multiple clicks if parents aren't expanded.)

### DIFF
--- a/src/client/js/Layouts/DefaultLayout/DefaultLayoutConfig.json
+++ b/src/client/js/Layouts/DefaultLayout/DefaultLayoutConfig.json
@@ -13,6 +13,12 @@
       "DEBUG_ONLY": false
     },
     {
+      "id": "ObjectBrowserPanel",
+      "panel": "ObjectBrowser/ObjectBrowserPanel",
+      "container": "east",
+      "DEBUG_ONLY": false
+    },
+    {
       "id": "VisualizerPanel",
       "panel": "Visualizer/VisualizerPanel",
       "container": "west",
@@ -22,12 +28,6 @@
       "id": "PartBrowserPanel",
       "panel": "PartBrowser/PartBrowserPanel",
       "container": "west",
-      "DEBUG_ONLY": false
-    },
-    {
-      "id": "ObjectBrowserPanel",
-      "panel": "ObjectBrowser/ObjectBrowserPanel",
-      "container": "east",
       "DEBUG_ONLY": false
     },
     {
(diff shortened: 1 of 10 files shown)


src/client/js/Layouts/DefaultLayout/DefaultLayoutConfig.json
```diff
@@ -13,6 +13,12 @@
       "DEBUG_ONLY": false
     },
     {
+      "id": "ObjectBrowserPanel",
```
kecso: is there a reason behind the changed order?

src/client/js/Panels/ModelEditor/ModelEditorControl.DiagramDesignerWidgetEventHandlers.js
```diff
@@ -1413,9 +1413,15 @@ define(['js/logger',
...
-            var gmeId = self._ComponentID2GmeID[id];
+            var gmeId = self._ComponentID2GmeID[id],
+                objDesc;
             if (self._GMEModels.indexOf(gmeId) > -1) {
-                selectedModels.push(self._getObjectDescriptor(gmeId));
+                objDesc = self._getObjectDescriptor(gmeId);
+                if (objDesc) {
+                    selectedModels.push(objDesc);
+                } else {
+                    self.logger.warn('_onAlignSelection, could not get objectDescriptor for a selectedIds', gmeId);
```
kecso: I believe the warning should be in singular instead of plural

src/client/js/Layouts/DefaultLayout/DefaultLayoutConfig.json
pmeijer: Yes, if it's not before the Visualizers the locate button will be to the right of the visualizer's button at initial load and when the visualizer is switched - the locate button will be to the left.
This way it's always to the left.

src/client/js/Panels/ModelEditor/ModelEditorControl.DiagramDesignerWidgetEventHandlers.js
pmeijer: You're right. Mind if I fix it elsewhere?

kecso: no, it is fine

PR #817, opened 2016-02-02
Fixes setPointer function of corediff.applyTreeDiff

kecso:
- the latest change in the patch application made some global variable unnecessary, but the setPointer function still had reference to those

### DIFF
--- a/test/common/core/corediff.spec.apply.js
+++ b/test/common/core/corediff.spec.apply.js
@@ -39,7 +39,7 @@ describe('corediff apply', function () {
         ]).nodeify(done);
     });
 
-    describe('apply', function () {
+    describe('basic', function () {
 
         beforeEach(function (done) {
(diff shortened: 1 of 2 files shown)


test/common/core/corediff.spec.apply.js
```diff
@@ -568,5 +566,268 @@ describe('corediff apply', function () {
...
+                        "cd891e7b-e2ea-e929-f6cd-9faf4f1fc045": true
+                    }
+                },
+                "guid": "e687d284-a04a-7cbc-93ed-ea941752d57a",
+                "oGuids": {
+                    "e687d284-a04a-7cbc-93ed-ea941752d57a": true
+                }
+            };
+
+            Q.nfcall(pointerCore.applyTreeDiff, pointerRootNode, patch)
```
pmeijer: You could use the coreQ instead..

PR #832, opened 2016-02-12
Performance improvements for core layers

lattmann:


src/common/core/coretree.js
```diff
@@ -695,29 +694,19 @@ define([
...
-                    result.push(keys[i]);
-                }
-            }
+            result = this.getRawKeys(node.data, predicate);
             return result;
         };
 
         this.getRawKeys = function (object, predicate) {
-            ASSERT(typeof predicate === 'undefined' || typeof predicate === 'function');
+            ASSERT(predicate === undefined || typeof predicate === 'function');
```
kecso: is this ASSERT really necessary?
I mean if the predicate is some 'not function' value the code will anyway throw an error when we try to use it as a function. (of course it is possible that the ASSERT gives us more information, but I doubt)

src/common/core/coretype.js
```diff
@@ -303,10 +304,14 @@ define([
         //</editor-fold>
 
         //<editor-fold=Modified Methods>
+        this.isValidNodeThrow = function (node) {
```
kecso: wouldn't it be better if the isValidNodeThrow functions would be 'helper' functions and not part of the class?

src/common/core/coretype.js
```diff
@@ -398,21 +403,22 @@ define([
             }, TASYNC.lift(children));
         };
 
+        function checkCollNames(node, draft) {
```
kecso: helper function should be positioned on the top of the file (driven by the editor-fold tags)

src/common/core/coretype.js
```diff
@@ -645,6 +651,60 @@ define([
             return Object.keys(merged);
         };
 
+        function getProperty(node, name) {
```
kecso: these are also all helper functions

PR #834, opened 2016-02-12
Fix json patch performance

kecso:
jsonPatcher converted the keys of the objects into array, which became too slow in case of high number of keys.

### DIFF
--- a/src/common/util/jsonPatcher.js
+++ b/src/common/util/jsonPatcher.js
@@ -7,7 +7,7 @@
  patch generation is specialized to the expected input form
  */
 
-define(['common/util/canon', 'underscore'], function (CANON, _) {
+define(['common/util/canon'], function (CANON) {
 
     function _strEncode(str) {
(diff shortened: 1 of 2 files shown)


src/common/util/jsonPatcher.js
```diff
@@ -21,33 +21,38 @@ define(['common/util/canon', 'underscore'], function (CANON, _) {
...
+                    }
                 }
 
                 //update
                 if (!noUpdate) {
-                    keys = _.intersection(tKeys, sKeys);
-                    for (i = 0; i < keys.length; i += 1) {
-                        if (CANON.stringify(source[keys[i]]) !== CANON.stringify(target[keys[i]])) {
-                            patch.push({op: 'replace', path: basePath + _strEncode(keys[i]), value: target[keys[i]]});
+                    for (i in target) {
```
pmeijer: In general I would avoid using `i` for a key iteration variable.

PR #835, opened 2016-02-12
Update executor readme with OSX and added dependent packages

lattmann:


src/server/middleware/executor/worker/package.json
```diff
@@ -3,8 +3,14 @@
...
     "agentkeepalive": "=2.0.3",
+    "balanced-match": "^0.3.0",
+    "concat-map": "0.0.1",
+    "inflight": "^1.0.4",
+    "inherits": "^2.0.1",
     "mime": ">=1.2.11",
     "minimatch": ">=0.3.0",
+    "once": "^1.3.3",
+    "path-is-absolute": "^1.0.0",
     "pkginfo": ">=0.3.0",
```
pmeijer: Are those packages used and weren't included before?
Or are they only used by the OS webkit?

lattmann: We had to install all of these packages on OSX in order to get the worker started. We did it on @VictorCoder123's machine.

PR #837, opened 2016-02-17
Fixes type error exception in jsonPatcher.

pmeijer:


pmeijer:
N.B. This needs to be cherry picked into v.1.5.x too

PR #844, opened 2016-02-18
Log output from executor-works to console.

pmeijer:


src/server/middleware/executor/worker/node_worker.classes.build.js
```diff
@@ -3640,7 +3699,8 @@ define('blob/BlobClient',[
                 hashes[filename] = hash;
 
                 if (err) {
-                    error += 'putFile error: ' + err.toString();
+                    error = err;
```
lattmann: It will return only with the last error, but that is fine, because we log all errors.

PR #869, opened 2016-03-02
Fixes error on connection removal from crosscut

kecso:
When a connection is displayed as a box on a crosscut it needs special handling during deletion.
That special procedure caused an exception as the id of the connection was removed earlier than the actual removal of the DOM element.


src/client/js/Panels/ControllerBase/DiagramDesignerWidgetMultiTabMemberListControllerBase.js
```diff
@@ -843,10 +843,9 @@ define(['js/logger',
 
                             //TODO: connection as box
                             //remove the box that represents this connections
```
pmeijer: Isn't this TODO covered by the change?

PR #873, opened 2016-03-02
Add DEMO tab in Projects Dialog.

pmeijer:
The tab is shown and the active one at start if authentication is enabled and the user is the guestAccount.

### DIFF
--- a/src/client/js/Dialogs/ProjectRepository/templates/ProjectRepositoryDialog.html
+++ b/src/client/js/Dialogs/ProjectRepository/templates/ProjectRepositoryDialog.html
@@ -4,7 +4,7 @@
             <div class="modal-header">
                 <button type="button" class="close" data-dismiss="modal">×</button>
                 <span class="title-container">
-                    <span class="project-repository-title">Commit History</span>
+                    <span class="project-repository-title">Project Repository</span>
                     <select class="form-control selector">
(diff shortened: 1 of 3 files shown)


src/client/js/Dialogs/ProjectRepository/templates/ProjectRepositoryDialog.html
```diff
@@ -4,7 +4,7 @@
             <div class="modal-header">
                 <button type="button" class="close" data-dismiss="modal">×</button>
                 <span class="title-container">
-                    <span class="project-repository-title">Commit History</span>
+                    <span class="project-repository-title">Project Repository</span>
```
pmeijer: I changed this back to the previous name...

src/client/js/Dialogs/ProjectRepository/templates/ProjectRepositoryDialog.html
```diff
@@ -4,7 +4,7 @@
...
                     <select class="form-control selector">
                         <optgroup class="history-selector-group group-project" label="Project">
                             <option value="$allBranchesAndTags">All branches & tags</option>
```
lattmann: I am going to merge this issue, but please change the `&` to `&amp;`.

PR #881, opened 2016-03-10
Added behavior for no activePanel set. Fixes #880

### LINKED ISSUE #880: ModelEditor crashes when using it without VisualizerPanel
The ModelEditor  crashes when used in a custom layout without the VisualizerPanel because the active panel is not set
https://github.com/webgme/webgme/blob/master/src/client/js/Panels/ModelEditor/ModelEditorControl.js#L1142

brollb:


brollb:
Double checked all the other visualizers and none had this problem

pmeijer:
Thanks Brian for taking care of this :+1:

PR #882, opened 2016-03-10
Fixes #852 Unable to create pointer after removal if it was defined in the base-class and itself.

### LINKED ISSUE #852: Unable to create pointer after removal if it was defined in the base-class and itself.
See comments in order.

### LINKED ISSUE #878: Attributes with prefix _
When I try to add an attribute starting with **underscore** character, WebGME doesn't create the attribute and doesn't give any warning or errors.

kecso:
The source of the issue was an unclear usage of data-driven and meta definition driven property names on the Meta editor.
- now we allow override of meta inherited meta definitions
- the set and aspect usage is clearly differentiated (sets cannot share names with aspects rigth now)
- it also fixes the issue of having '_' as prefix for attributes and other definition names ( Fixes #878)


pmeijer:
This seems to have failing tests..

PR #902, opened 2016-03-22
Set www.ogp.me metadata in index.html when requested.

pmeijer:
With this metadata e.g. facebook links will be decorated with metadata and images:

N.B. If behind a secure proxy, `gmeConfig.server.behindSecureProxy` need to be set to `true`.

FaceBook: (Using [URL-debugger](https://developers.facebook.com/tools/debug/og/object/).)
facebook

Skype:
skype

Slack:
slack


src/server/standalone.js
```diff
@@ -584,9 +585,33 @@ function StandAloneServer(gmeConfig) {
...
+        fs.readFile(indexHtmlPath, 'utf8', function (err, data) {
+            if (err) {
+                logger.error(err);
+                res.send(404);
+            } else {
+                res.contentType('text/html');
+                res.send(
+                    data.replace(/__URL__/g, url)
+                        .replace(/__IMAGE_URL__/g, imageUrl)
+                        .replace(/__TITLE__/g, projectId ? projectId.replace('+', '/') : 'WebGME')
```
ksmyth: Need to escape HTML entities here. Current code has a XSS vulnerability (e.g. set projectId to `"><script>...evil stuff...</script><a href="`)

pmeijer: Good point - thanks for pointing it out before it made the release.

Issue #905, opened 2016-03-22
Suppress auto visualizer look-up when switching panels in split view


pmeijer:

pmeijer:
Closed by #906

Issue #897, opened 2016-03-18
User experience issue : adding connector to a crosscut


phreed:
It is difficult to know if preexisting connectors exists between objects.
The main purpose of crosscut views is to establish a context in which 
connectors can be managed.

The current implementation is great for adding new connectors.
If there are two objects A and B and there is no connector between them;
then adding a connector is trivial. 
If there are more than one type of A-B connector that can be created;
a list of the different connector types is presented.

Assume there is already a A-B connector in another crosscut.
It may be that the A-B connector is relevant to the current crosscut. 
Presently, the A-B connector can be copied from the tree into the crosscut.
The difficulty is in identifying that the connection exists and which one it may be.
The means for identifying the other A-B connectors is an unpleasant experience.
The problem only manifests itself by examination of the "conpoisition" view
which, in some cases, will automatically include the new connectors.

Suggestion:
Enhance the list of connector types presented during creation 
to include all existing A-B connections in addition to the creation options.

pmeijer:
Do you use connector and connection interchangeably here?

pmeijer:
![goto](https://cloud.githubusercontent.com/assets/6518904/13968761/82f8cdb0-f04c-11e5-946e-abf7a2f4e278.PNG)

Fred is this sufficient. Clicking on the GoTo item will open up the existing connection. From there you can locate the node in the tree-browser. (Currently only the path is display since the connection may or may not be loaded at the time of the menu creation.)

pmeijer:
Currently if there is only one type of connection it is created automatically. With this change I would assume that the menu should at least pop up if there are existing connections. However would you like it to pop up even though there aren't any (and it that case display the only option 'Create connection...'?).

pmeijer:
One more thing, do you have a better phrase than 'GoTo'? 'Open'?

phreed:
Yes, that looks good. Open or use are both good verbs.

pmeijer:
Closed by #906

phreed:
There was more discussion on this issue.
The idea was not to go-to/open an existing connection but to select an existing connection for its inclusion in the cross-cut.

pmeijer:
Can you file a separate issue for that?

PR #906, opened 2016-03-22
Cross cut fixes #866 #897 #905

### LINKED ISSUE #866: Create a default crosscut
When the crosscut visualizer for an object is first opened there is no crosscut.
There is a message written on the backdrop saying that the operator needs to create one
but that is surprisingly easy to overlook.
(I completely misinterpreted what I was originally seeing. This no reflects the actual problem.)

pmeijer:


src/client/js/Panels/Crosscut/CrosscutController.js
```diff
@@ -226,6 +227,15 @@ define(['js/logger',
             crosscutsRegistry = GMEConcepts.getCrosscuts(memberListContainerID),
             len = crosscutsRegistry.length;
 
+        if (len === 0 && this._tabsHasBeenRequested === false) {
```
kecso: As this would automatically update the model, I think this should not be the default behavior, rather some option controlled.

PR #920, opened 2016-03-31
Send patch objects for all nodes and calc. changed nodes.

pmeijer:
With this implementation patch objects are always sent during a makeCommit and all core-objects are sent as such (not only the root).
On top of that the commitData also contains the changed nodes as dictionary, with keys update, partialUpdate, load and unload. They in turn contain the paths of the affected nodes. These are kept to a minimum and the client does some logic on top of that to figure out which nodes did have changes (load and unload are currently not used as events).

The breaking change is that a directly connected storage must make commits through the Project (unless there's no nodes in that commit). This is always possible since in order to get the core-objects in the first place the user needs a core which in turn needs a project.

The `emitCommittedCoreObjects` config has changed to `maxEmittedCoreObjects` which is a number (negative means not cap). N.B. this config only applies to newly created objects - not the patches (which are always sent).

### DIFF
--- a/src/client/js/client.js
+++ b/src/client/js/client.js
@@ -79,6 +79,7 @@ define([
                 loading: {
                     rootHash: null,
                     commitHash: null,
+                    changedNodes: null,
                     next: null
                 }
 
@@ -618,7 +619,7 @@ define([
                     if (!err && commitObj) {
                         logState('info', 'selectCommit loaded commit');
                         self.dispatchEvent(CONSTANTS.BRANCH_CHANGED, null);
-                        loading(commitObj.root, commitHash, function (err, aborted) {
+                        loading(commitObj.root, commitHash, null, function (err, aborted) {
                             if (err) {
                                 logger.error('loading returned error', commitObj.root, err);
                                 logState('error', 'selectCommit loading');
@@ -685,7 +686,7 @@ define([
                 self.dispatchEvent(CONSTANTS.REDO_AVAILABLE, canRedo());
 
                 logger.debug('loading commitHash, local?', commitHash, data.local);
-                loading(commitData.commitObject.root, commitHash, function (err, aborted) {
+                loading(commitData.commitObject.root, commitHash, commitData.changedNodes, function (err, aborted) {
                     if (err) {
                         logger.error('hashUpdateHandler invoked loading and it returned error',
                             commitData.commitObject.root, err);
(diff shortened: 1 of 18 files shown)


pmeijer:
@kecso before you merge let @lattmann have a look too.

src/common/storage/storageclasses/editorstorage.js
```diff
@@ -374,27 +374,42 @@ define([
...
             for (i = 0; i < keys.length; i += 1) {
-                if (coreObjects[keys[i]].oldHash &&
-                    coreObjects[keys[i]].newHash &&
-                    coreObjects[keys[i]].oldData &&
-                    coreObjects[keys[i]].newData) {
-                    //patch type object
+                if (UTIL.isPatchObjectFromCore(coreObjects[keys[i]])) {
+                    // Patch type object.
                     persistQueueElement[keys[i]] = coreObjects[keys[i]].newData;
+                    if (keys[i] === rootHash) {
```
lattmann: I would comment this line out, if the block is empty.

src/common/storage/storageclasses/editorstorage.js
```diff
@@ -374,27 +374,42 @@ define([
...
+                        //console.time('root-patch-computation');
+                    }
                     commitData.coreObjects[keys[i]] = UTIL.getPatchObject(coreObjects[keys[i]].oldData,
                         coreObjects[keys[i]].newData);
-
+                    if (keys[i] === rootHash) {
```
lattmann: I would comment this line out, if the block is empty.

src/common/util/jsonPatcher.js
```diff
@@ -182,8 +199,167 @@ define(['common/util/canon'], function (CANON) {
...
+        return RANDOM.isValidRelid(path.substring(1));
+    }
+
+    function _isGmePath(path) {
+        var relIds = path.split('/'),
+            result = false,
+            i;
+
+        for (i = 1; i < relIds.length; i += 1) {
+            if (RANDOM.isValidRelid(relIds[i]) === false) {
```
lattmann: maybe `return RANDOM.isValidRelid(relIds[i]) === true` instead?

src/common/util/jsonPatcher.js
```diff
@@ -182,8 +199,167 @@ define(['common/util/canon'], function (CANON) {
...
+        return false;
+    }
+
+    function _removeFromUpdates(res, gmePath) {
+        var updatesPath,
+            i;
+
+        updatesPath = Object.keys(res.update);
+        for (i = 0; i < updatesPath; i += 1) {
+            if (updatesPath[i].indexOf(gmePath) === 0) {
```
lattmann: A one line of comment would be great here. the indexOf operates on a string and not an array... this is like the startsWith function.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

src/common/util/jsonPatcher.js
```diff
@@ -182,8 +199,167 @@ define(['common/util/canon'], function (CANON) {
...
+                delete res.update[gmePath];
+            }
+        }
+
+        updatesPath = Object.keys(res.partialUpdate);
+        for (i = 0; i < updatesPath; i += 1) {
+            if (updatesPath[i].indexOf(gmePath) === 0) {
```
lattmann: Same here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

src/common/storage/util.js
```diff
@@ -52,6 +52,10 @@ define(['common/storage/constants', 'common/util/jsonPatcher'], function (CONSTA
 
             return patchObject;
         },
+        isPatchObjectFromCore: function (coreObj) {
```
pmeijer: This is bad naming..

src/common/util/jsonPatcher.js
pmeijer: All path-pieces need to pass the check and at least one needs to pass it too - so I can return after one truthy check. However I agree that it is a bit unclear from this code.

src/client/js/client.js
```diff
@@ -1138,17 +1139,85 @@ define([
...
+                }
+                //console.log('Update after meta considered', Object.keys(state.loading.changedNodes.update));
+
+                // 2. Add Root node
+                state.loading.changedNodes.update[ROOT_PATH] = true;
+
+                // 3. Account for loads and unloads.
+                for (loadUnloadPath in state.loading.changedNodes.load) {
+                    if (state.loading.changedNodes.load.hasOwnProperty(loadUnloadPath)) {
+                        pathPieces = loadUnloadPath.split('/');
```
kecso: I think, as a general rule of thumb, we should avoid path manipulation outside core (if it is unavoidable then we should provide some functions by the core API rather than implementing manipulation in different places)
Also it seems that the loaded and unloaded nodes are coming from the diff, but the diff cannot compute all the new nodes (as well as all the removed ones).

src/client/js/client.js
```diff
@@ -1138,17 +1139,85 @@ define([
...
+
+                //console.log('Update after loads and unloads considered',
+                //    Object.keys(state.loading.changedNodes.update));
+            }
+
             for (i in state.nodes) {
                 if (state.nodes.hasOwnProperty(i)) {
-                    if (newerNodes[i]) {
-                        if (newerNodes[i].hash !== state.nodes[i].hash && state.nodes[i].hash !== '') {
+                    if (newerNodes[i] && newerNodes[i].hash !== state.nodes[i].hash && state.nodes[i].hash !== '') {
```
kecso: do we still need to rely on the 'hash' field attached to the nodes?

test-karma/client/js/client.spec.js
```diff
@@ -2723,7 +2724,10 @@ describe('GME client', function () {
                 }
 
                 if (testState === 'checking') {
-                    expect(events).to.have.length(8);
+                    events.forEach(function (e) {
+                        console.log(e);
```
kecso: console.log
- either test the events, or comment out the logging

test/server/storage/safestorage.spec.js
```diff
@@ -1124,32 +1197,125 @@ describe('SafeStorage', function () {
...
+                    .then(function (fcoNode) {
+                        var persisted;
+                        importResult.core.createNode({
+                            parent: rootNode,
+                            base: fcoNode
+                        });
+                        persisted = importResult.core.persist(rootNode);
+                        expect(Object.keys(persisted.objects).length).to.equal(2);
+                        newRootHash = persisted.rootHash;
+                        console.log(persisted);
```
kecso: console.log residue

src/client/js/client.js
pmeijer: Good point. Is there a path divider constants or maybe even a method?
Does this apply to the patcher too?

src/client/js/client.js
pmeijer: I'm not sure, it's still being set to an integer. I haven't touched that part though.

PR #923, opened 2016-04-07
Adds feature to export commit-queue and apply it to a branch.

pmeijer:


bower.json
```diff
@@ -45,6 +45,7 @@
     "require-css": "0.1.8",
     "q": "1.4.1",
     "codemirror": "5.12.0",
-    "jquery.gritter": "1.7.4"
+    "jquery.gritter": "1.7.4",
+    "remarkable-bootstrap-notify": "3.1.3"
```
pmeijer: I ended up not using this, but I think it will be useful to have as a compliment to notifications in the footers.

src/client/js/Utils/LocalStorageManager.js
```diff
@@ -0,0 +1,76 @@
+/*globals define, $, window, WebGMEGlobal*/
+/**
+ * TODO: This is work in progress and currently not used anywhere.
```
pmeijer: You can disregard this file, it's not used. However I'd like to keep it checked in as it has some example of the $.notify and if we end up using the local storage (either for the commit-queue or user-state) - we have a placeholder for it.

src/client/js/Panels/Header/ProjectNavigatorController.js
```diff
@@ -801,14 +811,44 @@ define([
...
+                    }
+
+                    var dialog = new ApplyCommitQueueDialog(self.gmeClient, WebGMEGlobal.gmeConfig, branches);
+                    dialog.show(data, function (commitQueue, opts) {
+                        self.gmeClient.applyCommitQueue(commitQueue, opts, function (err, result) {
+                            if (err) {
+                                self.logger.error(err);
+                            }
+
+                            console.log(result);
```
kecso: console.log

src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
```diff
@@ -3,14 +3,16 @@
...
  */
 
 define([
     'js/logger',
     'js/Controls/DropDownMenu',
     'js/Controls/PopoverBox',
     'js/Dialogs/Merge/MergeDialog',
-    'js/Constants'
+    'js/Constants',
+    'bootstrap-notify'
```
kecso: does it not return with an object?
if we not use it, either comment it out or remove or add an 'unused' object to the function's parameter list

pmeijer: It adds itself onto $.notify. I will remove the line though

PR #925, opened 2016-04-08
Add library support to WebGME

kecso:
This is the first phase of the introduction of libraries into WebGME.
- Libraries are a core concept
- Any full project can behave as a library
- There is no inherent connection among libraries of your project
- The libraries in a project are read-only
- The source of a library can be any project package or intra-deployment URL of a project (meaning that project from other deployments must be copied over first)
- Libraries introducing the FQN (fully qualified name), that is a namespace like concatenation of the library contaners of the node and its own name
- You can update/rename/remove your libraries, but the imported libraries should always have unique names

Other changes also included:
- The 'view meta rules' functionality of the MetaEditor became read-only (so now you only allowed to select the definition)
- When the cardinality rules of containment reach their limit, the part browser 'grays out' those elements that should not be instentiated anymore


src/client/js/Dialogs/Import/ImportDialog.js
```diff
@@ -150,6 +150,7 @@ define(['js/Loader/LoaderCircles',
 
                     var checkResult = WebGMEGlobal.Client.checkImport(parsedJSONFileContent,
                         CORE_CONSTANTS.EXPORT_TYPE_LIBRARY);
+                    checkResult = null;
```
pmeijer: Is this deprecated?

src/client/js/Dialogs/LibraryManager/removeConfirmDialog.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * @author kecso / https://github.com/kecso
```
pmeijer: Why can you reuse ConfirmDelete?

src/client/js/Dialogs/Import/ImportDialog.js
kecso: Yes, or at least it needs serious refactoring.
On the other hand, this whole file became deprecated, so I believe this change is just some early residue...

src/client/js/Dialogs/LibraryManager/updateDialog.js
```diff
@@ -0,0 +1,157 @@
...
+             CORE_CONSTANTS,
+             AssetWidget,
+             UTILS,
+             REGEXP,
+             updateDialogTemplate) {
+
+    'use strict';
+
+    var UpdateDialog,
+        MAX_FILE_SIZE = 100000000;
```
pmeijer: Is this coming from the blob? Extract it as a constant..

src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
```diff
@@ -741,16 +909,15 @@ define(['js/logger',
                 toggled: {
                     hideConnections: false,
                     hideAbstracts: false,
-                    hideLeaves: false
+                    hideLeaves: false,
+                    hideLibraries: false
```
pmeijer: Update `webgme\config\componentsGenericUIDefaults.json`

src/client/js/Panels/PropertyEditor/PropertyEditorPanelController.js
```diff
@@ -216,6 +217,17 @@ define(['js/logger',
                     isCommon: true,
                     readOnly: true
                 };
+
+                if (cNode.isLibraryElement()) {
+                    propList[' GUIDl'] = {
+                        name: 'GUID[library]',
```
pmeijer: GUID (library)

src/client/js/Widgets/TreeBrowser/TreeBrowserWidget.js
```diff
@@ -453,25 +456,38 @@ define(['js/logger',
             }
         }
 
-        if (objDescriptor.isAbstract !== node.data.isAbstract) {
+        if (objDescriptor.hasOwnProperty('isAbstract') && objDescriptor.isAbstract !== node.data.isAbstract) {
```
pmeijer: nice

src/client/js/Widgets/TreeBrowser/TreeBrowserWidget.js
```diff
@@ -729,24 +744,30 @@ define(['js/logger',
...
                     name: 'Select node',
                     callback: function (/*key, options*/) {
                         self.onMakeNodeSelected.call(self, node.key);
                     },
                     icon: false
+                },
+                open: { // The "select (aka double-click)" menu item
+                    name: 'Open in',
+                    items: {
+                        visualizer: {
```
pmeijer: nice

src/client/js/Widgets/TreeBrowser/TreeBrowserWidget.js
```diff
@@ -987,6 +1006,20 @@ define(['js/logger',
             addToggleBtn('hideLeaves', 'leaf nodes', '<i class="fa gme-atom"/>');
         }
 
+        if (typeof options.hideLibraries === 'boolean') {
+            self._currentFilters.hideLibraries = options.hideLibraries;
+            self._filterFunctions.push(function (node) {
+                // Filter out connection nodes.
```
pmeijer: library nodes

src/client/js/client/gmeNodeSetter.js
```diff
@@ -5,7 +5,7 @@
  */
 define([], function () {
     'use strict';
-    function gmeNodeSetter(logger, state, saveRoot, storeNode) {
+    function gmeNodeSetter(logger, state, saveRoot, storeNode, sendError) {
```
pmeijer: printCoreError, why change?

src/client/js/client/gmeNodeSetter.js
```diff
@@ -28,32 +28,56 @@ define([], function () {
         }
 
         function setAttributes(path, name, value, msg) {
+            var error;
+
             if (state.core && state.nodes[path] && typeof state.nodes[path].node === 'object') {
-                state.core.setAttribute(state.nodes[path].node, name, value);
+                error = state.core.setAttribute(state.nodes[path].node, name, value);
+                if (error instanceof Error) {
```
pmeijer: These are recoverable from core, clients point of view right?

src/client/js/client/libraries.js
```diff
@@ -0,0 +1,126 @@
+/*globals define*/
+/*jshint browser: true*/
+/**
+ * @author kecso / https://github.com/kecso
+ */
+define(['js/Constants'], function (CONSTANTS) {
+    'use strict';
+    function gmeLibraries(logger, state, storage, saveFunction) {
```
pmeijer: nodeSetters have a function called storeNode if they are the same I would use the same name.

src/client/js/client/libraries.js
```diff
@@ -0,0 +1,126 @@
...
+            state.core.removeLibrary(state.nodes[CONSTANTS.PROJECT_ROOT_ID].node, libraryName);
+            saveFunction('removeLibrary(' + libraryName + ')');
+        }
+
+        function renameLibrary(oldName, newName) {
+            state.core.renameLibrary(state.nodes[CONSTANTS.PROJECT_ROOT_ID].node, oldName, newName);
+            saveFunction('renameLibrary(' + oldName + ',' + newName + ')');
+        }
+
+        function followLibrary(libraryRootId) {
```
pmeijer: openLibraryOriginInNewWindow

src/common/blob/util.js
```diff
@@ -72,20 +72,21 @@ define(['q', './BlobMetadata'], function (Q, BlobMetadata) {
      * @param callback
      * @returns {*}
      */
-    function addAssetsFromExportedProject(logger, blobClient, mainMetadata, callback) {
+    function addAssetsFromExportedProject(logger, blobClient, mainMetadata) {
```
pmeijer: A general rule of thumb is that public functions should take a callback when they return a callback

src/common/blob/util.js
```diff
@@ -98,12 +99,126 @@ define(['q', './BlobMetadata'], function (Q, BlobMetadata) {
...
                 }
-                return projectFileHash;
+                return deferred.resolve(projectFileHash);
+            })
+            .catch(deferred.reject);
+
+        return deferred.promise;
+    }
+
+    function _gatherFilesFromMetadataHashRec(logger, blobClient, metadata, assetHash, artifact) {
```
pmeijer: ...same here

src/common/blob/util.js
```diff
@@ -98,12 +99,126 @@ define(['q', './BlobMetadata'], function (Q, BlobMetadata) {
...
+
+            Q.ninvoke(artifact, 'addMetadataHash', filenameContent, assetHash, metadata.size)
+                .then(function () {
+                    return Q.ninvoke(artifact, 'addFile', filenameMetadata, JSON.stringify(metadata));
+                })
+                .then(function () {
+                    deferred.resolve();
+                })
+                .catch(function (err) {
+                    err = err instanceof Error ? err : new Error(err);
```
pmeijer: BlobClient always return with Errors.

src/common/blob/util.js
```diff
@@ -98,12 +99,126 @@ define(['q', './BlobMetadata'], function (Q, BlobMetadata) {
...
+                }))
+                .then(function () {
+                    // Finally add the .metadata for the complex object.
+                    return Q.ninvoke(artifact, 'addFile', filenameMetadata, JSON.stringify(metadata));
+                })
+                .then(function () {
+                    deferred.resolve();
+                })
+                .catch(function (err) {
+                    deferred.reject(new Error(err));
```
pmeijer: same here (this will make the Error.toString() print Error twice)

src/common/blob/util.js
```diff
@@ -98,12 +99,126 @@ define(['q', './BlobMetadata'], function (Q, BlobMetadata) {
...
+                });
+        } else {
+            deferred.reject(new Error('Unsupported content type', metadata.contentType));
```
pmeijer: + metadata.contentType

src/common/blob/util.js
```diff
@@ -98,12 +99,126 @@ define(['q', './BlobMetadata'], function (Q, BlobMetadata) {
...
+     *
+     * @param {GmeLogger} logger
+     * @param {BlobClient} blobClient
+     * @param {object} jsonExport
+     * @param {boolean} addAssets
+     * @param callback
+     * @returns {*}
+     */
+    function buildProjectPackage(logger, blobClient, jsonExport, addAssets, callback) {
+        console.log('exporting.', addAssets);
```
pmeijer: console.log

src/common/blob/util.js
```diff
@@ -98,12 +99,126 @@ define(['q', './BlobMetadata'], function (Q, BlobMetadata) {
...
+        var artie = blobClient.createArtifact(jsonExport.projectId +
+                '_' + (jsonExport.branchName || jsonExport.commitHash)),
+            assets = jsonExport.hashes.assets || [];
+
+        artie.descriptor.name = jsonExport.projectId +
+            '_' + (jsonExport.branchName || jsonExport.commitHash) + '.webgmeX';
+        console.log('exporting:', assets);
```
pmeijer: console.log

src/common/blob/util.js
```diff
@@ -98,12 +99,126 @@ define(['q', './BlobMetadata'], function (Q, BlobMetadata) {
...
+                var error,
+                    i;
+
+                for (i = 0; i < result.length; i += 1) {
+                    if (result[i].state === 'rejected') {
+                        error = result[i].reason instanceof Error ? result[i].reason : new Error(result[i].reason);
+                        logger.debug('Gathering returned with error', assets[i], error);
+                        if (error.message.indexOf('Another content with the same name was already added.') === -1) {
+                            //some real error
+                            logger.error('Error for gathering asset failed:', assets[i], error);
```
pmeijer: No need to log the error (since there's no extra info I would let the caller log it).

src/common/blob/util.js
```diff
@@ -98,12 +99,126 @@ define(['q', './BlobMetadata'], function (Q, BlobMetadata) {
...
+                        return _gatherFilesFromMetadataHashRec(logger, blobClient, metadata, assetHash, artie);
+                    });
+            }))
+            .then(function (result) {
+                var error,
+                    i;
+
+                for (i = 0; i < result.length; i += 1) {
+                    if (result[i].state === 'rejected') {
+                        error = result[i].reason instanceof Error ? result[i].reason : new Error(result[i].reason);
```
pmeijer: Again blob-client returns Errors.

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+
+        //library function TODO checking everything and adding all new functions
+
+        /**
+         * It adds a project as library to your project by copying it over. The library will be a node
+         * with the given name directly under your project's ROOT. It becomes a read-only portion of your project.
+         * You will only be able to manipulate it with library functions, but cannot edit the individual nodes inside.
+         * However you will be able to instantiate or copy the nodes into other places of your project. Every node
+         * that was part of the META in the originating project becomes part of your project's meta.
+         * @param {module:Core~Node} node - any regular node in your project.
```
pmeijer: This is good doc (some tweaks can be made) I would however extract the conceptual things to a wiki-page. And assume the core user know what a library is.

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         * @param {module:Core~Node} node - any node in your project.
+         *
+         * @return {module:Core~Node} - Returns the acting FCO of your project.
+         *
+         * @func
+         */
+        this.getFCO = core.getFCO;
+
+        /**
+         * Returns if the node in question is a library root or not.
```
pmeijer: Returns true if the node in question is a library root.

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         *
+         * @return {bool} - Returns true if your node is a library root (even if it is embedded in other library),
+         * false otherwise.
+         *
+         * @func
+         */
+        this.isLibraryRoot = core.isLibraryRoot;
+
+        /**
+         * Returns if the node in question is a library element or not.
```
pmeijer: same

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         *
+         * @return {bool} - Returns true if your node is a library element, false otherwise.
+         *
+         * @func
+         */
+        this.isLibraryElement = core.isLibraryElement;
+
+        /**
+         * Returns the fully qualified name of any node, which is the list of its namespaces separated
+         * by comma and followed by the name of the node.
```
pmeijer: by dot

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         *
+         * @param {module:Core~Node} node - the node in question.
+         *
+         * @return {string} - Returns the fully qualified name of the node (like: 'library.myName').
```
pmeijer: name of the node.

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         * @param {string} newName - the new name of the project.
+         *
+         * @func
+         */
+        this.renameLibrary = core.renameLibrary;
+
+        /**
+         * Returns the info associated with the library.
+         *
+         * @param {module:Core~Node} node - any node in the project.
```
pmeijer: What will happen if it's a regular node or library element?

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         */
+        this.getLibraryInfo = core.getLibraryInfo;
+
+        /**
+         * Returns the core node of the given library.
+         *
+         * @param {module:Core~Node} node - any node in the project.
+         * @param {string} name - the name of the library.
+         *
+         * @return {module:Core~Node | null} - Returns the library root node or null, if invalid library was queried.
```
pmeijer: What is an invalid library?

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         *
+         * @func
+         */
+        this.getLibraryRoot = core.getLibraryRoot;
+
+        /**
+         * Returns all the Meta nodes found in the library.
```
pmeijer: replace `found` with `within the given library node`.

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         *
+         * @param {module:Core~Node} node - any node of your project.
```
pmeijer: Again any node?

src/common/core/coreQ.js
```diff
@@ -191,6 +191,28 @@ define(['common/core/core', 'q'], function (Core, Q) {
...
+
+            return deferred.promise.nodeify(callback);
+        };
+
+        var updateLibraryOrg = this.updateLibrary;
+        this.updateLibrary = function (node, name, updatedLibraryRootHash, libraryInfo, updateInstructions, callback) {
+            var deferred = Q.defer();
+            updateLibraryOrg(node, name, updatedLibraryRootHash, libraryInfo, updateInstructions, function (err, result) {
+                if (err) {
+                    deferred.reject(err);
```
pmeijer: is this always Error?

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+
+            var logger = options.logger,
+                self = this,
+                key;
+
+            for (key in innerCore) {
+                this[key] = innerCore[key];
+            }
+
+            // We need to remove guid related functions so they will not appear on the public API.
```
pmeijer: Doesn't core.js take care of this?

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                    libraryRoots = {},
+                    path,
+                    roots,
+                    i,
+                    name;
+
+                for (path in allMetaNodes) {
+                    roots = getAllLibraryRoots(allMetaNodes[path]);
+                    for (i = 0; i < roots.length; i += 1) {
+                        name = getLibraryName(roots[i]) + self.getAttribute(roots[i], 'name');
```
pmeijer: If we use libraries the name has a special meaning, maybe we should consider not being able to remove it from the UI at least.

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+            }
+
+            function getRootOfLibrary(node, name) {
+                var root = self.getRoot(node);
+
+                return root.libraryRoots[name];
+            }
+
+            function getLibraryRoot(node) {
+                while (node) {
```
pmeijer: In `getLibraryName` you use parent and aren't changing the argument.

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                    parent = self.getParent(parent);
+                }
+
+                return name;
+            }
+
+            function getRootOfLibrary(node, name) {
+                var root = self.getRoot(node);
+
+                return root.libraryRoots[name];
```
pmeijer: Could be a one-liner.

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+            delete this.getDeductedGuid;
+
+            logger.debug('initialized LibraryCore');
+
+            //<editor-fold=Helper Functions>
+            function getLibraryRootsInfo(node) {
+                var allMetaNodes = self.getAllMetaNodes(node),
+                    libraryRoots = {},
+                    path,
+                    roots,
```
pmeijer: Kutakinte

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                    }
+                }
+
+                return libraryRoots;
+            }
+
+            function getLibraryName(node) {
+                ASSERT(self.isValidNode(node));
+                var parent,
+                    name = '';
```
pmeijer: Minor thing, try to put variable declarations with assignments at the top.

src/client/js/client/gmeNodeSetter.js
kecso: Yes. Basically the error only means, that although the type of the parameters are probably right, the circumstances do not allow the execution of the function.

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                        return node;
+                    }
+                    node = self.getParent(node);
+                }
+
+                return null;
+            }
+
+            function getAllLibraryRoots(node) {
+                var roots = [];
```
pmeijer: parent variable

src/client/js/client/libraries.js
kecso: during the library functions we do not receive 'new' nodes that needs to be stored, that is why it is not used here (just the saveFunction which I will rename to saveRoot to make the similarity clear)

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                    return root;
+                }, innerCore.loadRoot(hash));
+            };
+
+            this.createNode = function (parameters) {
+                var node;
+
+                if (parameters && parameters.parent &&
+                    (self.isLibraryRoot(parameters.parent) || self.isLibraryElement(parameters.parent))) {
+                    return new Error('cannot create new node inside a library');
```
pmeijer: Start sentence with upper case and end it with a dot.
"cannot" could mean that it could do it under other conditions -> "Not allowed to ..."

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                }
+
+                if (parameters && parameters.base && self.isLibraryRoot(parameters.base)) {
+                    return new Error('cannot instantiate a library root');
```
pmeijer: The jsdocs should be update with @returns {undefined|Error}

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         *
+         * @return {object} - Returns the information object, stored alongside the library (that basically
+         * carries metaData about the library).
+         *
+         * @func
+         */
+        this.getLibraryInfo = core.getLibraryInfo;
+
+        /**
+         * Returns the core node of the given library.
```
pmeijer: 'the core node' -> 'library root'

src/common/core/metacachecore.js
```diff
@@ -140,6 +140,18 @@ define([
 
                 return {};
             };
+            
+            this.getFCO = function(node){
+                var root = self.getRoot(node),
+                    key;
+                
+                for(key in root.metaNodes){
```
pmeijer: spacing

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+
+                            if (newLibraryNodes.length > 0) {
+                                //connect the FCO as base of libraryFCO, but be sure to remove the nullPtr
+
+                                libraryFCO = self.getBaseRoot(newLibraryNodes[0]);
+                                innerCore.deletePointer(libraryFCO, 'base');
+                                innerCore.setBase(libraryFCO, FCO);
+
+                                //adding new nodes to the global META
+                                console.log('NNP', newNodePaths, inMeta);
```
pmeijer: consule.lug

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+
+            this.updateLibrary = function (node, name, updatedLibraryRootHash, libraryInfo, updateInstructions) {
+                var logs = {added: {}, updated: {}, moved: {}, removed: {}},
+                    root = self.getRoot(node),
+                    libraryRoot = getRootOfLibrary(root, name),
+                    relid,
+                    FCO = self.getFCO(root),
+                    oldLibraryRootHash;
+
+                if (!libraryRoot) { //do nothing if not valid library
```
pmeijer: why not Error here?

src/common/core/core.js
kecso: this just needs any node, so it can lookup the library info, the only 'meaningful' input here is the name of the library

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                        //refreshing libraryInfo
+                        root.libraryRoots[name] = newLibraryRoot;
+                    }, self.loadSubTree(newLibraryRoot));
+                }, self.loadChild(root, libraryRelid));
+            };
+
+            this.updateLibrary = function (node, name, updatedLibraryRootHash, libraryInfo, updateInstructions) {
+                var logs = {added: {}, updated: {}, moved: {}, removed: {}},
+                    root = self.getRoot(node),
+                    libraryRoot = getRootOfLibrary(root, name),
```
pmeijer: There are a lot of roots here. libraryRoot, RootOfLibrary and root. Why not getLibraryRootByName.

src/common/core/core.js
kecso: yes. I know that is could seem odd, but as these informaitons are always cached, I believe it is more useful if you can query with any arbitrary node (you just need the name of the library)

src/common/core/coreQ.js
kecso: yes

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                                    if (newNodePaths.indexOf(self.getPath(newLibraryNodes[i])) !== -1 &&
+                                        inMeta.indexOf(self.getPath(newLibraryNodes[i])) !== -1) {
+                                        innerCore.addMember(root, CONSTANTS.META_SET_NAME, newLibraryNodes[i]);
+                                    }
+                                }
+                            }
+
+                            root.libraryRoots[name] = newLibraryRoot;
+
+                            return logs;
```
pmeijer: Where does logs come from?

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                        }, self.loadSubTree(newLibraryRoot));
```
pmeijer: What will happen with a huge library here? Is it necessary to load them all at once?

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+            this.updateLibrary = function (node, name, updatedLibraryRootHash, libraryInfo, updateInstructions) {
+                var logs = {added: {}, updated: {}, moved: {}, removed: {}},
+                    root = self.getRoot(node),
+                    libraryRoot = getRootOfLibrary(root, name),
+                    relid,
+                    FCO = self.getFCO(root),
+                    oldLibraryRootHash;
+
+                if (!libraryRoot) { //do nothing if not valid library
+                    return logs;
```
pmeijer: Why not Error here?

src/common/core/librarycore.js
kecso: what do you mean?
you cannot simply 'delete' libraries
you cannot clear their name (and you need to give them one when you add them)
also the 'name' attribute cannot be removed from FCO (though that is somewhat different thing)

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+
+            };
+
+            this.removeLibrary = function (node, name) {
+                ASSERT(self.isValidNode(node));
+                var root = self.getRoot(node);
+
+                ASSERT(typeof name === 'string' &&
+                    name.indexOf(CONSTANTS.NAMESPACE_SEPARATOR) === -1 &&
+                    root.libraryRoots[name]);
```
pmeijer: These will throw if not libraries...

src/common/core/librarycore.js
```diff
@@ -0,0 +1,785 @@
...
+                        libraryNodes[path] = allNodes[path];
+                    }
+                }
+
+                return libraryNodes;
+            };
+
+            this.getLibraryInfo = function (node, name) {
+                var libroot = getRootOfLibrary(node, name);
+                return self.getAttribute(libroot, '_libraryInfo');
```
pmeijer: This will throw too if not a library root.

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         *
+         * @param {module:Core~Node} node - any node of your project.
+         * @param {string} name - name of your library.
+         *
+         * @return {module:Core~Node[]} - Returns an array of core nodes that are part of your meta from
+         * the given library.
+         *
+         * @func
+         */
+        this.getLibraryNodes = core.getLibraryNodes;
```
pmeijer: getLibraryMetaNodes

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         * Returns the core node of the given library.
+         *
+         * @param {module:Core~Node} node - any node in the project.
+         * @param {string} name - the name of the library.
+         *
+         * @return {module:Core~Node | null} - Returns the library root node or null, if invalid library was queried.
+         *
+         * @func
+         */
+        this.getLibraryRoot = core.getLibraryRoot;
```
pmeijer: getLibraryRootByName

src/common/core/librarycore.js
kecso: Yeah, that is the wrong 'pattern'. I've changed that.

src/common/core/core.js
```diff
@@ -1659,6 +1663,173 @@ define([
...
+         *
+         * @param {module:Core~Node} node - the node in question.
+         *
+         * @return {string} - Returns the fully qualified name of the node (like: 'library.myName').
+         *
+         * @func
+         */
+        this.getFullyQualifiedName = core.getFullyQualifiedName;
+
+        /**
```
pmeijer: For the methods down here think about the naming..
- getLibraryInfo
- removeLibrary
- getLibraryGuid
- renameLibrary
- getLibraryNodes

Some take a name and some take a rootNode. It's not clear which require what.

src/common/core/users/meta.js
```diff
@@ -21,10 +24,11 @@ define([], function () {
             _saveFunc(message);
         }
 
-        function initialize(core, nodes, save) {
+        function initialize(core, nodes, save, error) {
```
pmeijer: Can you initialize two meta?
error is not a good name for an error-handler.

src/common/core/users/meta.js
```diff
@@ -83,32 +87,52 @@ define([], function () {
 
                 //children
                 if (meta.children && meta.children.items && meta.children.items.length > 0) {
-                    _core.setChildrenMetaLimits(node, meta.children.min, meta.children.max);
+                    error = _core.setChildrenMetaLimits(node, meta.children.min, meta.children.max);
```
pmeijer: this is not the same as the passed in function..

src/server/worker/workerrequests.js
```diff
@@ -850,7 +1374,11 @@ function WorkerRequests(mainLogger, gmeConfig) {
         checkConstraints: checkConstraints,
         // This is exposed for unit tests..
         _addZippedExportToBlob: _addZippedExportToBlob,
-        reassignGuids: reassignGuids
+        reassignGuids: reassignGuids,
+        saveProjectIntoFile: save,
+        importProjectFromFile: load,
```
pmeijer: rename save and load

src/common/core/librarycore.js
pmeijer: I meant the name attribute defined on FCO.

src/common/core/librarycore.js
kecso: getLibraryRootByName could be confused with getLibraryRoot, because both function would need a node to get access to the containment hierarchy. I know it is not ideal, but I hope that at least the public API function names are straigthforward

src/common/core/librarycore.js
kecso: At some point we need all of the nodes (again we cannot figure out the move operation without getting all nodes). You are right, that we will probably need some optimization. I will check with Hakan's big project, and if it has issues we will search for alternatives during phase 2.

src/common/core/users/meta.js
kecso: please explain

src/common/core/users/meta.js
kecso: in theory, yes you can but probably it should be changed...

src/common/core/core.js
kecso: we can iterate on the names, for now I made them consistent in the sense that they always get an arbitrary node as an input, and then a library name

PR #926, opened 2016-04-09
Plugin metadata and support for plugin Icons. Closes #898 too.

### LINKED ISSUE #898: Requiring text in plugin causes tests to fail
When testing a plugin that uses the "text!" prefix the tests fail with error 

`Tried loading "text" at ... then tried node's require("text") and it traile failed with error: Error: Cannot find module 'text' to equal null `

pmeijer:
Plugins now support metadata in form of a json. This way info about the plugin can be loaded without loading the plugin and instantiate it. The `./metadata.json` holds static info about the plugin, `id`, `name`, `description`, `pluginConfig` etc. The new data is `icon` that can either hold a relative path to an image file put in an ``-element (see [Wikipedia](https://en.wikipedia.org/wiki/Comparison_of_web_browsers#Image_format_support) for browser support) and/or a css class (the generic ui uses [Bootstrap Glyphicons](http://bootstrapdocs.com/v3.3.6/docs/components/#glyphicons-glyphs) and [Font Awesome Icons (4.1.0)](https://fortawesome.github.io/Font-Awesome/icons/)).

In this version **old plugins are still supported** but this will be phase out in later versions. The breaking change is related to the rest API route `'/plugins'`.

New calls:
+ GET - `'/plugins/metadata'`
+ GET - `'/plugin/:pluginId/metadata'`

Changed calls:

+ GET - `'/plugins/:pluginId/<path>'` -> `'/plugin/:pluginId/<path>'`

##Example
Below is an example. See any of the coreplugins for more details.
```
{
  "id": "ConstraintEvaluator",
  "name": "Constraint Evaluator",
  "version": "1.7.0",
  "description": "Plugin for developers to test, evaluate and add custom constraints.",
  "icon": {
    "class": "glyphicon glyphicon-ok-circle",
    "src": "",
  },
  "disableServerSideExecution": false,
  "disableBrowserSideExecution": false,
  "configStructure": [
    {
      "name": "mode",
      "displayName": "Execution mode",
      "description": "Select if Constraints.js should be evaluated, generated or populated in model.",
      "value": "EvaluateConstraints",
      "valueType": "string",
      "valueItems": [
        "EvaluateConstraints",
        "GenerateConstraints",
        "PopulateFromConstraints"
      ]
    },
    {
      "name": "clear",
      "displayName": "Clear existing constraints",
      "description": "Clear all existing constraints when populating model with constraints.",
      "value": false,
      "valueType": "boolean"
    }
  ]
}
```


src/client/js/Dialogs/PluginConfig/PluginConfigDialog.js
```diff
@@ -74,8 +76,21 @@ define(['js/Controls/PropertyGrid/PropertyGridWidgetManager',
         this._divContainer = this._dialog.find('.modal-body');
 
         this._saveConfigurationCb = this._dialog.find('.save-configuration');
+        this._modalHeader = this._dialog.find('.modal-header');
 
-        this._title = this._dialog.find('.modal-title');
+        if (typeof this._pluginMetadata.icon === 'string') {
```
kecso: wouldn't it be more clear to have two fields (like iconClass and iconSrc) than assuming that if it contains dot, then it is a path?

lattmann: What would be the order/precedence? Would we apply the iconClass on the img tag, too?

kecso: I would say that it is probably not a bad idea to allow extra classes for an icon file as it would give even more options. That would also makes it clear that there is no need for precedence.

src/plugin/PluginBase.js
```diff
@@ -127,8 +120,7 @@ define([
      * @returns {string}
      */
     PluginBase.prototype.getName = function () {
-        throw new Error('implement this function in the derived class - getting type automatically is a bad idea,' +
-            'when the js scripts are minified names are useless.');
+        return this.pluginMetadata.name;
```
kecso: if the default 'null' metada is not overridden, this cause an exception.
Do we intended to have this change of behavior? (or should we still send some nice error instead)

test/server/api/plugin/plugin.spec.js
```diff
@@ -83,8 +83,86 @@ describe('REST API', function () {
...
+                        ]);
+
+                        done();
+                    });
+            });
+
+            it('should 404 for non-existing plugin /api/plugin/DoesNotExist/metadata', function (done) {
+                agent.get(server.getUrl() + '/api/v1/plugin/DoesNotExist/metadata')
+                    .end(function (err, res) {
+                        console.log(res.body);
```
kecso: console.log

test/server/api/plugin/plugin.spec.js
```diff
@@ -83,8 +83,86 @@ describe('REST API', function () {
...
+                        expect(res.status).equal(404, err);
+                        done();
+                    });
+            });
+
+            it('should 404 for plugin with no metadata /api/plugin/PluginForked/metadata', function (done) {
+                agent.get(server.getUrl() + '/api/v1/plugin/PluginForked/metadata')
+                    .end(function (err, res) {
+                        console.log(res.body);
```
kecso: console.log

src/client/js/Dialogs/PluginConfig/PluginConfigDialog.js
pmeijer: Agree, adding icon: {src: .. , class: ...}

lattmann: :+1:

finger563:
minor point: the example above needs to have the `,` at the end of `"src": "",` removed to be used as an actual example (otherwise `JSON.parse` will fail)

pmeijer:
@finger563 It's been corrected. Thanks!

PR #935, opened 2016-04-11
Fixes library element ordering in ObjectBrowser

kecso:
- children should be ordered when a new node is added
- only libraryRoots should be on top of the child order

### DIFF
--- a/src/client/js/Widgets/TreeBrowser/TreeBrowserWidget.js
+++ b/src/client/js/Widgets/TreeBrowser/TreeBrowserWidget.js
@@ -293,6 +293,7 @@ define(['js/logger',
                 isConnection: objDescriptors[i].isConnection,
                 isAbstract: objDescriptors[i].isAbstract,
                 isLibrary: objDescriptors[i].isLibrary,
+                isLibraryRoot: objDescriptors[i].isLibraryRoot,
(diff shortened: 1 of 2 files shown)


src/client/js/Widgets/TreeBrowser/TreeBrowserWidget.js
```diff
@@ -364,10 +365,12 @@ define(['js/logger',
             isConnection: objDescriptor.isConnection,
             isAbstract: objDescriptor.isAbstract,
             isLibrary: objDescriptor.isLibrary,
+            isLibraryRoot: objDescriptor.isLibraryRoot,
             libraryInfo: objDescriptor.libraryInfo,
             metaType: objDescriptor.metaType
         }, beforeNode);
 
+        this.sortChildren(parentNode, false);
```
pmeijer: Doesn't this mean that it sorts the children even though there wasn't any changes?

kecso: you added a new child, that is the change, so it should order them

pmeijer: Ah, I read the diff wrong.. Thought it was in the same methods as the other changes.. My mistake!

PR #953, opened 2016-04-15
Make it possible to run under systemd with socket activation

phreed:
http://0pointer.de/blog/projects/socket-activation.html
https://www.freedesktop.org/software/systemd/man/systemd.socket.html
https://developer.atlassian.com/blog/2015/03/docker-systemd-socket-activation/

There are two relevant commits in this branch.
The first one is abandoned as I had some problems with the node-systemd package.
It is retained for reference.
The attached zip file contains sample systemd config files.

[init_systemd.zip](https://github.com/webgme/webgme/files/221402/init_systemd.zip)


phreed:
I am making a cleaner branch.

PR #956, opened 2016-04-18
prepare webgme to start as a service with "systemd" the new "init"

phreed:
http://0pointer.de/blog/projects/socket-activation.html
https://www.freedesktop.org/software/systemd/man/systemd.socket.html
https://developer.atlassian.com/blog/2015/03/docker-systemd-socket-activation/

This makes use of https://www.npmjs.com/package/systemd-socket to adopt the systemd socket.
https://github.com/herzi/systemd-socket

The attached zip file contains sample systemd config files.

[init_systemd.zip](https://github.com/webgme/webgme/files/224751/init_systemd.zip)

There is another package which does similar things but uses monkey-patching.
https://github.com/rubenv/node-systemd


src/server/standalone.js
```diff
@@ -194,9 +195,20 @@ function StandAloneServer(gmeConfig) {
...
+        // should be considered to shut the service down if it
+        // is not getting much use.
+
+        // https://github.com/herzi/systemd-socket
+        // https://nodejs.org/api/http.html#http_server_listen_handle_callback
+        __httpServer.listen(systemdSocket(0) || gmeConfig.server.port, function () {
         __httpServer.listen(gmeConfig.server.port, function () {
             // Note: the listening function does not return with an error, errors are handled by the error event
-            logger.debug('Http server is listening on ', {metadata: {port: gmeConfig.server.port}});
+            logger.error('Http server is listening on ',
```
pmeijer: It's loggin an error here..

pmeijer:
The tests are failing with syntax errors in standalone. Btw you can make more commits to your branch with worrying about the history in the master branch in webgme; Github provides a "squash" option before merging (this came out last week or so).

src/server/standalone.js
```diff
@@ -194,9 +195,19 @@ function StandAloneServer(gmeConfig) {
...
 
-        __httpServer.listen(gmeConfig.server.port, function () {
+        // something link
+        // https://github.com/rubenv/node-autoquit
+        // should be considered to shut the service down if it
+        // is not getting much use.
+
+        // https://github.com/herzi/systemd-socket
+        // https://nodejs.org/api/http.html#http_server_listen_handle_callback
+        __httpServer.listen(systemdSocket(0) || gmeConfig.server.port, function () {
```
brollb: It looks like `systemdSocket(0)` returns the port number... If so, why don't you just add this to your config files and set `gmeConfig.server.port` to this value?

phreed: Are you suggesting moving the systemd code to one of the 'config/config.*.js' files?
Yes, that would probably work fine.

brollb: yeah, that's what I mean. I don't like the idea of introducing a systemd specific dependency...

This should be doable similar to how you can allow users to set the port w/ env variables: https://github.com/dfst/deepforge/blob/master/config/config.base.js#L10

phreed: Sorry I took so long to get back to this.
The problem with the approach you advocate is that systemSocket(0) does not return a number.
systemSocket(0) returns an object of the form '{fd: 3}'.
This should be simple enough to fix.
The problem is in the validator, it does not like an argument of the form '{fd: 3}'.
I hacked webgme/config/validator.js:194
// assertNumber('config.server.port', config.server.port);
This allows the file descriptor adoption.

pmeijer:
This is resolved in #1206

PR #977, opened 2016-04-25
Enable webgmex usage in import/export bin scripts

kecso:
import.js and export.js works with webgmex packages only!
_v1 versions still available until 2.0.0 for older versions.

PR should be squashed and cherry-picked into 1.7.x branch


src/bin/export.js
```diff
@@ -1,264 +1,262 @@
...
+    return runInternal(program);
+};
+
+/**
+ * Entry point for usage as a module.
+ *
+ * @param {Dictionary<String,String>} params
+ * @return {undefined}
+ */
+run = function (params) {
```
pmeijer: Why not `function run(params) {`?
Navigating to run will bring one to the declaration at the top the way it is now..

src/bin/export.js
```diff
@@ -1,264 +1,262 @@
...
+        program = new Command(),
+        syntaxFailure = false;
+
+    gmeConfig = require(path.join(process.cwd(), 'config'));
+    logger = webgme.Logger.create('gme:bin:export', gmeConfig.bin.log, false);
+    blobClient = new BlobClient(gmeConfig, logger.fork('BlobClient'));
+    webgme.addToRequireJsPaths(gmeConfig);
+    mainDeferred = Q.defer();
+    program
+        .version('0.2.0')
```
pmeijer: Bump the version to 1.7.2.

src/bin/export.js
```diff
@@ -1,264 +1,262 @@
...
+            }
+
+            if (params.user) {
+                project.setUser(params.user);
+            }
+
+            return storageUtils.getProjectJson(project, projectParams);
+        })
+        .then(function (jsonExport) {
+            return Q.ninvoke(blobUtil,'buildProjectPackage',logger, blobClient, jsonExport, params.noAssets !== true);
```
pmeijer: Doesn't `blobUtil.buildProjectPackage` return a promise?

src/bin/export.js
kecso: noted. 
I removed run and runRaw as they are not needed anymore.
We could also clear out runInternal, but for now it can stay and it uses the `function runInternal(` form of definition.

PR #980, opened 2016-04-27
Plugin execution

pmeijer:
The breaking change is that plugins are now required to have an associated `metadata.json` and the execution of browser plugins (once configured) are now handled by the client.

This PR deals with the following:

- Enables plugins to be executed under a namespace, i.e. the `this.META` within the plugin can be configured to contain only nodes under a given namespace (within a library). This allows the meta helpers function on PluginBase to adapt to the namespace.

- Plugin metadata introduces new key `requiresWriteAccess` - which prohibits execution of a plugin if the invoker does not have write access to the project. (The UI disables these plugins and the plugin manager throws an error if requirement not met.)

- Plugin configuration parameters are also extend with `requiresWriteAccess`. They are handled in the same way as above. (`readOnly` parameters are ensured to not be overwritten by the users too.)

- Execution of plugins in the browser share the same code base as for the server execution. Additionally there is no need to use the InterpreterManager anymore (it is still there but only deals with configuring plugins) - instead the client has a method `runBrowserPlugin` with the same parameters as `runServerPlugin`.


src/bin/run_plugin.js
```diff
@@ -113,6 +119,12 @@ main = function (argv, callback) {
             logger.info('Project is opened.');
             project = project_;
 
+            return gmeAuth.getProjectAuthorizationByUserId(params.username, params.projectId);
+        })
+        .then(function (access) {
+            logger.info('User has the following writes to the project: ', access);
```
kecso: writes -> rights

src/plugin/managerbase.js
```diff
@@ -0,0 +1,471 @@
...
+         *
+         * @param {string} pluginId
+         * @param {object} pluginConfig - configuration for the plugin.
+         * @param {object} context
+         * @param {string} [context.branchName] - name of branch that should be updated
+         * @param {string} [context.commitHash=<%brashHash%>] - commit from which to start the plugin.
+         * @param {ProjectInterface} [context.project=project] - project instance if different from the one passed in.
+         * @param {string} [context.activeNode=''] - path to active node
+         * @param {string[]} [context.activeSelection=[]] - paths to selected nodes.
+         * @param {string} [context.namespace=''] - used namespace during execution ('' is the root).
```
kecso: root seems a bit misleading ('' represents all namespaces ? - maybe)

src/plugin/managerbase.js
```diff
@@ -0,0 +1,471 @@
...
+                    pluginContext.activeNode = activeNode;
+                    self.logger.debug('activeNode loaded');
+                    // Load active selection
+                    return self.loadNodesByPath(pluginContext, context.activeSelection || []);
+                })
+                .then(function (activeSelection) {
+                    pluginContext.activeSelection = activeSelection;
+                    self.logger.debug('activeSelection loaded');
+                    // Load meta nodes
+                    var metaIds = pluginContext.core.getMemberPaths(pluginContext.rootNode, 'MetaAspectSet');
```
kecso: this should be replaced with the getAllMetaNodes synchronous call (probably faster...)

src/plugin/managerbase.js
```diff
@@ -0,0 +1,471 @@
...
+                    }
+
+                    if (context.namespace) {
+                        if (libraryNames.indexOf(context.namespace) === -1) {
+                            throw new Error('Given namespace does not exist among the available: "' +
+                                libraryNames + '".');
+                        }
+
+                        for (i = 0; i < metaNodes.length; i += 1) {
+                            nodeNamespace = pluginContext.core.getNamespace(metaNodes[i]);
```
kecso: wouldn't it be the same to use simply the getFullyQualifiedName and cut the context namespace from that?
(fewer variables and no 'name' constant in the code)
or if the name can have dots, I would still use the FQN and the namespace so no 'name' constant is used...

pmeijer: That could lead to collisions. If a meta node, in a project with a library named `A`, starts with `A.` - it would appear in the `A` namespace.

PR #983, opened 2016-04-28
Fixes #982 Gathering collection information for a node inside a complex instantiation crashes the core

### LINKED ISSUE #982: Gathering collection information for a node inside a complex instantiation crashes the core
Given the following scenario:
- type/node A
- type/node B that has a child of type A (that is an instance of the node A)
- type/node C
- type/node S that has a child of type B (that is an instance of node B)
- type/node D that has a child of type C and a child of type A and the C typed child refers to A typed child with pointer 'ref'
- type/node V that has a child of type S and a child of type D

In this setup, if we create an instance of type V (let's call it V'), then when we ask the names of the collections of its A typed grandchild, we should receive the list ['ref']. Right now, such question ends in a crash inside the Core.

kecso:


src/common/core/constants.js
```diff
@@ -16,6 +16,7 @@ define([], function () {
         MEMBER_RELATION: 'member',
         BASE_POINTER: 'base',
         PATH_SEP: '/',
+        MUTABLE_PROPERTY: '_mutable',
```
pmeijer: Nice

src/common/core/coretype.js
```diff
@@ -177,117 +177,110 @@ define([
             return node;
         }
 
+        /**
```
pmeijer: Would you expect this new look-up to be more computationally expensive?

kecso: It should not.
Actually it can be that it will be faster after the change.
Originally, the getChild operation was used to get the overlay information and that can consume resources (as it could end up creating empty core wrapper objects). Now we use the getProperty instead which always returns a json-like object and use less resources.
Node-wise we check roughly the same amount of nodes, or even fewer as now every recursion step gathers the correct 'instantiation sub-tree'.

pmeijer: If these are often called in conjunction, we could consider having a combined method (to me it looks like they are doing the same iterations).

pmeijer: Sounds good to me :)

PR #987, opened 2016-05-04
Fixes #963 Adding and removing library node to main meta sheet removes the node from the global meta.

### LINKED ISSUE #955: Better feedback when trying to edit/add a constraint to a library node.
Related to #963

### LINKED ISSUE #963: Adding and removing library node to main meta sheet removes the node from the global meta.
It should only be removed it from the sheet.

Nodes should also have their fully qualified name when present on the meta sheet - ideally some read-only indication too.

Related to #955

kecso:
- now MetaDecorator uses FQN
- the library elements are 'read-only' but can be put onto any sheet
- when removing library elements, they will be removed without question as they cannot loose their meta information
Fixes #955
- constraints now cannot be added/changed for library nodes

### DIFF
--- a/src/client/decorators/MetaDecorator/DiagramDesigner/MetaDecorator.DiagramDesignerWidget.js
+++ b/src/client/decorators/MetaDecorator/DiagramDesigner/MetaDecorator.DiagramDesignerWidget.js
@@ -38,7 +38,8 @@ define([
     var MetaDecoratorDiagramDesignerWidget,
         DECORATOR_ID = 'MetaDecorator',
         ABSTRACT_CLASS = 'abstract',
-        TEXT_META_EDIT_BTN_BASE = $('<i class="glyphicon glyphicon-cog text-meta"/>');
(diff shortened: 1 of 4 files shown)


src/client/decorators/MetaDecorator/DiagramDesigner/MetaDecorator.DiagramDesignerWidget.js
```diff
@@ -76,23 +77,34 @@ define([
...
     //jshint camelcase: false
     MetaDecoratorDiagramDesignerWidget.prototype.on_addTo = function () {
-        var self = this;
+        var self = this,
+            node = self._control._client.getNode(self._metaInfo[CONSTANTS.GME_ID]),
+            belongsToLibrary = false;
+
+        if (node && (node.isLibraryElement() || node.isLibraryRoot())) {
+            belongsToLibrary = true;
+            self._inLibrary = true;
```
pmeijer: Aren't these two the same?

PR #998, opened 2016-05-19
Fixed dockerfile so it builds and runs the default configuration

kecso:


Dockerfile
```diff
@@ -1,73 +1,64 @@
-FROM ubuntu:14.04
```
pmeijer: Add instructions.

Dockerfile
```diff
@@ -1,73 +1,64 @@
...
 RUN echo smallfiles = true >> /etc/mongodb.conf
 
-RUN mkdir -p /project
-ADD package.json /project/package.json
+WORKDIR /usr/app
 
-# mount the current project workspace under /project inside the container
-ADD . /project
+# copy app source
+# TODO we should take .dockerignore into use
```
pmeijer: Copy entire directory

PR #1002, opened 2016-05-24
Feature display connected users

pmeijer:


src/client/debug.html
```diff
@@ -11,10 +11,10 @@
     <meta property="og:site_name" content="WebGME"/>
     <meta property="og:description" content="Web-based Generic Modeling Environment"/>
 
-    <!--<script>-->
-        <!--document.write('<script src="http://' + (location.host || 'localhost').split(':')[0] +-->
-                <!--':35729/livereload.js?snipver=1"></' + 'script>')-->
-    <!--</script>-->
+    <script>
```
kecso: is it ok if it becomes 'default'?
I though it need some configuring with the livereload

test/server/storage/websocket.spec.js
```diff
@@ -217,6 +217,274 @@ describe('WebSocket', function () {
...
+                            join: true,
+                            projectId: projectEmitNotification,
+                            branchName: 'master'
+                        })
+                    ]);
+                })
+                .then(function () {
+                    var deferred = Q.defer();
+                    receiver.socket.on(CONSTANTS.NOTIFICATION, function (data) {
+                        console.log(data);
```
kecso: console.log

PR #1010, opened 2016-05-31
Make project authorization (read, write, delete, create) a replaceable module.

pmeijer:
gmeAuth is broken down into three components:
- `authorizer` - handles project authorization regarding read, write, delete and create.
- `metadataStorage` - stores metadata about projects (not replaceable).
- `gmeAuth` - Handles authentication and user management for default authorizer.

The rest calls regarding get /user(/:settings) ensures a users is in the db (user is already authenticated at that point).


src/server/middleware/auth/authorizerbase.js
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * @type {{PROJECT: string, USER: string}}
+ */
+AuthorizerBase.ENTITY_TYPES = ENTITY_TYPES;
+
+/**
+ *
+ * @param {object} params
+ * @param {object} params.collection - Mongo collection to default database.
```
kecso: Wouldn't it be better to have these parameters only in the proper portions of gmeConfig?
I mean if a different implementation would use a completely different implementation of 'start', then we would have to change other portion of the code as well.

kecso: If it would be hairy to 'create' the collection object similar to gmeAuth inside the default authorizer, than it can stay as we can go around it (just ignoring it...), but eventually I think we should avoid this 'constraint'.

pmeijer: I agree that it's not the nicest solution, but since we have the connection initiated from gmeAuth (and will use it for the project metadata and to store user settings regardless of the chosen authorizer) it might as well be passed into the authorizer. It's up to any implementation to use it or not. The reason I made start/stop asynchronous is to give allow for custom authorizers to initiate connections or what ever they need to do.

PR #1012, opened 2016-06-05
Fixes #1007 Library update fails with "object does not exist".

### LINKED ISSUE #1007: Library update fails with "object does not exist"
When trying to update a library from a webgmex file, it fails and says that `Error: object does not exist #9855825a7d06d9fcd19491b146f1de1b00accc0a(…)`

kecso:
In case of an imported project, the library cannot be found in its original form in the database, so the information of the nodes in the library should be collected from the library form.


pmeijer:
Before you merge this, can you add a test failing with the previous implementation?

pmeijer:
:+1:

PR #1019, opened 2016-06-10
Traverse function have been implemented.

kecso:


src/common/core/core.js
```diff
@@ -1961,6 +1961,26 @@ define([
...
+         *
+         * @param {module:Core~Node} root - the root node of the sub-tree that needs to be traversed.
+         * @param {object} options - parameters to control the traversing.
+         * @param {bool} [options.excludeRoot = false] - controls whether the root should be excluded from visit.
+         * @param {integer} [options.maxParallelLoad = 10]- the maximum number of parallel loads allowed.
+         * @param {bool} [options.blockingVisit = true] - controls if the loading should wait for the visit
+         * function to return.
+         * @param {integer} [options.speed = 10] - the frequency in milliseconds
+         * to check for the next task in the traverse.
+         * @param {function()} visitFn - the synchronous visitation function that will be called for
```
pmeijer: Are we limiting this to synchronous work?

kecso: Technically we cannot limit that, we just expect the visit function to be synchronous.
Majority of the exchange formats can be handled in a synchronous way (in regards to the visit function) and we can extend it if there is a real need (by adding another version of the traverse that deals with an asynchronous visit function).
The function footprint is less complex this way.

pmeijer: So did we agree on making the `visitFn` asynchronous? I will remove me as assignee for now..

kecso: yes.
I will assign you once I have done with the changes.

test/common/core/coretreeloader.spec.js
```diff
@@ -0,0 +1,182 @@
...
+                expect(checkArray).to.have.length(118);
+                expect(checkArray.indexOf('/-1')).not.to.equal(-1);
+                expect(checkArray.indexOf('')).not.to.equal(-1);
+            })
+            .nodeify(done);
+    });
+
+    it('should traverse without blocking visit', function (done) {
+        var checkArray = [];
+        core.traverse(rootNode, {blockingVisit: false}, function (node, next) {
```
pmeijer: This option is deprecated right?

test/common/core/coretreeloader.spec.js
```diff
@@ -0,0 +1,182 @@
...
+                expect(checkArray).to.have.length(117);
+                expect(checkArray.indexOf('/-1')).not.to.equal(-1);
+                expect(checkArray.indexOf('')).to.equal(-1);
+            })
+            .nodeify(done);
+    });
+
+    it('should traverse with minimal settings', function (done) {
+        var checkArray = [];
+        core.traverse(rootNode, {maxParallelLoad: 1, speed: 10, blockingVisit: true}, function (node, next) {
```
pmeijer: These too

PR #1043, opened 2016-06-27
Add in user-management-page for dealing with users.

pmeijer:
This adds `/post/registration`.
Removes the redirect at `login` (handled by the client instead).

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "jszip": "2.5.0",
     "method-override": "2.3.5",
     "mime": "1.3.4",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "superagent": "1.7.2",
     "underscore": "1.8.3",
     "unzip": "0.1.11",
+    "user-management-page": "webgme/user-management-page#master",
```
pmeijer: N.B. This will point to the 0.1.0 release before webgme is released.

PR #1047, opened 2016-06-29
Generator for rest router

pmeijer:
This PR also includes:
- RestRouters now have start/stop methods that are invoked at start/stop of the server.
- External rest-routers are mounted directly at their given path (`/rest/external/` is not prepended) N.B. This is the breaking change!
- Also note that the old RestComponent is removed (the format is supported but will log a warning).
- Start and stop things in correct order when starting/stopping server.


src/plugin/coreplugins/RestRouterGenerator/Templates/combine_templates.js
```diff
@@ -0,0 +1,84 @@
+/*jshint node:true*/
```
kecso: why do we use the templates this way?
wouldn't it be more simple to require them as text?

src/server/middleware/ExampleRestComponent.js
```diff
@@ -1,5 +1,7 @@
 /*jshint node:true*/
```
kecso: I think we can remove this file completely.

pmeijer: I'll remove it

src/plugin/coreplugins/RestRouterGenerator/Templates/combine_templates.js
pmeijer: Yes you're right and I tried it too but ran into requirejs issues.
Looked closer at it and turns out if the name contains `.js` it throws the following error at the tests. 'Error: ENOENT: no such file or directory, open 'C:\GIT\webgme\plugin\RestRouterGenerator\RestRouterGenerator\RestRouter.js.ejs`... So I will just rename it.

PR #1054, opened 2016-07-07
Fixes #1033 Error when merging branches - GUID Mismatch

### LINKED ISSUE #1033: Error when merging branches - GUID Mismatch
Tried to merge into existing branch and got the following error. The only thing that should have changed are a couple of attributes, as far as I know.

screen shot 2016-06-23 at 6 20 12 pm

kecso:
When one of the differences of the merge is empty, the merge should still gracefully generate the patch.
Also during diff generation the merge should avoid of creating difference information based on 'null' data.

### DIFF
--- a/test/common/core/corediff.spec.js
+++ b/test/common/core/corediff.spec.js
@@ -148,9 +148,7 @@ describe('core diff', function () {
                     if (err) {
                         return done(err);
                     }
-
-                    console.log(JSON.stringify(diff, null, 2));
-                    console.log(JSON.stringify(patch, null, 2));
+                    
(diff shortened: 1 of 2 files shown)


test/common/core/corediff.spec.js
```diff
@@ -140,6 +149,8 @@ describe('core diff', function () {
                         return done(err);
                     }
 
+                    console.log(JSON.stringify(diff, null, 2));
+                    console.log(JSON.stringify(patch, null, 2));
```
pmeijer: logs..

PR #1061, opened 2016-07-15
Fix issues with asymmetric merge behavior 

kecso:
When generating the conflicts, the merge should not be asymmetrical in terms of input diffs.


pmeijer:
Tests are failing so this will have to wait till 2.3.0

PR #1062, opened 2016-07-15
Add missing rest api call for transfer projects

patrickkerrypei:


pmeijer:
Looks good Patrick, thanks

PR #1065, opened 2016-07-18
Closes #1064 do not set the Content-Length from blob client when running in the browser

### LINKED ISSUE #1064: Error on artifact.save()
It doesn't cause any noticeable problems but it is annoying to have errors in the dev console (especially when demoing)

```
Refused to set unsafe header "Content-Length"    superagent.js:1019 
```

pmeijer:


ksmyth:
What was the issue with this? The trouble is that the server does not know when all the data has been sent. Maybe we should research this more

ksmyth:
Ah, I see that browsers will complain. But what about node? And the executor worker

pmeijer:
Thanks for the input.
Yeah so the browsers do not even allow you to set the header. But you're saying that node processes depend on setting the header?

pmeijer:
Ok, I see these are needed for the keep-alive connections.
It looks like superagent sets this header automatically for node processes too. Do you know of any cases where it wouldn't work?

pmeijer:
Since node-js doesn't complain about the header and it worked fine before. I will only remove it when running under the browser.

PR #1068, opened 2016-07-20
Fixed getArtifact promise error handling. Fixes #1067

### LINKED ISSUE #1067: BlobClient getArtifact promise "callback is not a function"
Using the blobClient's `getArtifact` method (w/o a callback and expecting a promise), an error will result in an exception being thrown

The problem is https://github.com/webgme/webgme/blob/master/src/common/blob/BlobClient.js#L516

brollb:


pmeijer:
Thanks Brian

PR #1075, opened 2016-07-22
Auto merge when making commits

pmeijer:
When the server receives an outdated commit into a branch - instead of directly returning the FORKED status, it will trigger a worker that attempts to merge the commit into the branch. If it succeeds, the client will get the MERGED status and the storage on the client side will handle it appropriately. If the merging returns with conflicts or with an unexpected error - the client will receive the FORKED status.

Although the clients do not have to wait for the the merges, there is a risk that the client builds up a large commit-queue.

### DIFF
--- a/src/common/core/users/merge.js
+++ b/src/common/core/users/merge.js
@@ -19,39 +19,33 @@ define([
     'use strict';
 
     function save(parameters, callback) {
-        var deferred = Q.defer(),
-            persisted = parameters.core.persist(parameters.root),
+        var persisted = parameters.core.persist(parameters.root),
             newRootHash;
 
         if (persisted.hasOwnProperty('objects') === false || Object.keys(persisted.objects).length === 0) {
             parameters.logger.warn('empty patch was inserted - not making commit');
-            deferred.resolve({
+            return Q({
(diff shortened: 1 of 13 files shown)


src/common/core/users/merge.js
```diff
@@ -271,6 +272,10 @@ define([
                 return doMerge();
             })
             .then(function () {
+                var ms = Date.now() - startTime,
+                    min = Math.floor(ms/1000/60),
+                    sec = (ms/1000) % 60;
+                console.log('Merge exec time', min, 'min', sec, 'sec');
```
kecso: console.log

test/server/storage/websocket.spec.js
```diff
@@ -1136,46 +1425,89 @@ describe('WebSocket', function () {
...
+                socketSend,
+                branchName = 'b2',
+                data = {
+                    projectId: projectName2Id('WebSocketTest_BRANCH_HASH_UPDATED'),
+                    branchName: branchName,
+                    join: true
+                },
+                newBranchHash,
+                eventHandler = function (resultData) {
+                    console.log('eHandle');
```
kecso: console.log

PR #1086, opened 2016-07-30
Added ORIGIN_URL env on ExecutorWorker creation. Fixes #1085

### LINKED ISSUE #1085: detect originating webgme url from executor script
When running a script from within an executor, there should be a way to determine the originating webgme url (for me, this is relevant bc I want to use the BlobClient to upload intermediate results but there are certainly other use cases).

There is the `config.json` provided to the worker but it can contain multiple webgme urls and not necessarily the originating url.

I think it would be good to set an environment variable when spawning the child process which contains the originating webgme url.

brollb:


pmeijer:
I will make the node_worker build and merge this. Thanks brian.

PR #1118, opened 2016-08-18
Fixes #1106 Importing project fails: "Error: Invalid argument, data.projectId is not a string.(…)"

### LINKED ISSUE #1106: Importing project fails: "Error: Invalid argument, data.projectId is not a string.(…)"
When loading the root page and creating a new project, I get the following error:
```
logger.js:66 Cannot select project Error: Error: Invalid argument, data.projectId is not a string.(…)
```

kecso:
The used zip library uses string based errors instead of Error objects, so we have to convert them so the client would receive the correct error.


brollb:
This is better as we are now getting the error in the dev console. In the future, it would be good to notify the user using the web ui

PR #1125, opened 2016-08-23
Closes #1094 Requirejs build

### LINKED ISSUE #1094: Speed up build
it looks like building the `classes.js` files take one of the longest - could this be set up as a `prepublish`?

### LINKED ISSUE #1113: Error in JsDoc on BlobClient
https://github.com/webgme/webgme/blob/28476e01ec2106e476181acca2d51f02455f7fde/src/common/blob/BlobClient.js#L560
I think you mean...
* @param {int} [bytes] - File size in bytes.

pmeijer:
This PR adds the previous postinstall script as a prepublish script. The postinstall is only used (calling prepublish) if the distribution files do not exist.

Additionally it splits up the distribution into two separate files, one for libraries and one for the source code. The reason is that non-minified files are used for the libraries - the previous source-map became to large for chrome to load. Now the library files are minified from non-minified files.

Epiceditor is now a bower-component and bundled with the other libraries.

It also fixes a minor documentation typo and closes #1113.


utils/build/dist/build.js
```diff
@@ -92,20 +90,95 @@ var requirejs = require('requirejs'),
...
+        return deferred.promise;
+    }
+
+    return Q.all([
+        callOptimizer(config),
+        callOptimizer(cssConfig),
+        callOptimizer(libConfig)
+    ])
+        .then(function (result) {
+            console.log('Build time', (Date.now() - start) / 1000, 's');
```
kecso: console.log - though in this case it could probably stay

pmeijer: It can stay

PR #1149, opened 2016-09-06
Core performance enhancements

pmeijer:
1. Use children-relids as an object when existence is checked.
2. Cached children should be stored in an object rather than array.
3. Reuse code to enable better optimization from engine (and to prepare for future changes in overlay handling).

These changes gives around 10% faster querying of models (loading, getting all attribute names and values).

### DIFF
--- a/src/common/core/coretree.js
+++ b/src/common/core/coretree.js
@@ -41,52 +41,73 @@ define([
         // ------- memory management
 
         function __detachChildren(node) {
-            ASSERT(node.children instanceof Array && node.age >= CONSTANTS.MAX_AGE - 1);
+            //ASSERT(node.children instanceof Array && node.age >= CONSTANTS.MAX_AGE - 1);
 
             var children = node.children;
             node.children = null;
             node.age = CONSTANTS.MAX_AGE;
 
-            for (var i = 0; i < children.length; ++i) {
-                __detachChildren(children[i]);
+            for (var child in children) {
+                __detachChildren(children[child]);
             }
         }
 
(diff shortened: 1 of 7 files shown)


src/common/core/coretree.js
```diff
@@ -38,55 +37,74 @@ define([
...
+            var root,
+                i;
             if (++ticks >= CONSTANTS.MAX_TICKS) {
                 ticks = 0;
-                __ageNodes(roots);
+                for (i = 0; i < roots.length; i += 1) {
+                    root = roots[i];
+                    ASSERT(root.age < CONSTANTS.MAX_AGE);
+                    if (++root.age >= CONSTANTS.MAX_AGE) {
+                        roots.splice(i, 1);
```
kecso: if you want to splice inside the loop, then you should use decreasing iterator not to skip any elements

src/common/core/coretype.js
```diff
@@ -79,7 +79,7 @@ define([
                 }
             }
             //normal child - as every node should have a base, it is normally mean a direct child of the ROOT
-            if (self.getChildrenRelids(node).indexOf(relid) === -1) {
+            if (self.getChildrenRelids(node, true).hasOwnProperty(relid) === false) {
```
kecso: wouldn't the ```self.getChildrenRelids(node,true)[relid] !== true``` be faster?

src/common/core/coretree.js
pmeijer: Good catch!

PR #1163, opened 2016-09-13
Only check if AHEAD for server plugins. Fixes #1162

### LINKED ISSUE #1162: Should be able to execute client side plugin while AHEAD
When executing a server side plugin, it makes sense to block the use if the client is AHEAD since the server doesn't have the same state as the client (which is the target state of the plugin execution). However, this doesn't hold if the plugin is to be executed on the client as the client has all the necessary state for the plugin.

brollb:


pmeijer:
If the plugin forks locally from the client - it will proceed to make commits on it's own. Since the client is AHEAD there is no guarantee that the commit the plugin started from would have been persisted at the time the plugin's commit arrives.

src/client/js/Utils/InterpreterManager.js
```diff
@@ -96,14 +96,15 @@ define([
...
                 // Before executing the plugin - make sure the client is in SYNC.
                 // (If not plugin could be executed on the server on a commitHash that does not exist).
-                if (self._client.getBranchStatus() &&
-                    self._client.getBranchStatus() !== self._client.CONSTANTS.BRANCH_STATUS.SYNC) {
-                    callback(self.getPluginErrorResult(metadata.id, 'Not allowed to invoke plugin while local branch' +
-                        ' is AHEAD or PULLING changes from server.', startTime));
-                    return;
-                }
-
                 if (globalConfig.runOnServer === true) {
```
pmeijer: globalConfig.runOnServer === true || metadata.writeAccessRequired === true

brollb: Updated this to make it more legible and to address this issue in the latest commit!

pmeijer: Not a fan of the double negation..

pmeijer: I will merge it though. Thanks for the contribution!

PR #1166, opened 2016-09-15
Added rest router unit test generation. Fixes #1146

### LINKED ISSUE #1146: Generate unit test file(s) for rest routers

brollb:
when writing the test, I assume people are using the webgme-cli and added some comments on what will need to be changed if they are not using the cli.

I figured the benefits of doing things like looking up the mount point in the webgme-setup.json outweigh the inconvenience of having to change it otherwise.

### DIFF
--- a/src/plugin/coreplugins/RestRouterGenerator/RestRouterTest.ejs
+++ b/src/plugin/coreplugins/RestRouterGenerator/RestRouterTest.ejs
@@ -3,7 +3,9 @@ var testFixture = require('../../globals'),  // TODO: May need to change this if
     expect = testFixture.expect,
     gmeConfig = testFixture.getGmeConfig(),
     server = testFixture.WebGME.standaloneServer(gmeConfig),
-    mntPt = require('../../webgme-setup.json').components.routers['<%= restRouterName %>'].mount,
+    // TODO: If not using webgme-cli, replace the mntPt w/ the desired mount point
+    // If using the webgme-cli, this will look up the mount point for the given router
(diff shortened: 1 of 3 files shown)


src/plugin/coreplugins/RestRouterGenerator/RestRouterTest.ejs
```diff
@@ -0,0 +1,67 @@
+var testFixture = require('../../globals'),  // TODO: May need to change this if not created from webgme-cli
```
pmeijer: Would you mind moving these within the describe?
It might not be that big of a deal.. But if you have a lot of test files and a recursive look-up and only want to run some tests using `only` it will load all these modules.
(Keeping the testFixture here is fine since all tests practically use that)

src/plugin/coreplugins/RestRouterGenerator/RestRouterTest.ejs
```diff
@@ -0,0 +1,67 @@
...
+    superagent = testFixture.superagent,
+    expect = testFixture.expect,
+    gmeConfig = testFixture.getGmeConfig(),
+    server = testFixture.WebGME.standaloneServer(gmeConfig),
+    // TODO: If not using webgme-cli, replace the mntPt w/ the desired mount point
+    // If using the webgme-cli, this will look up the mount point for the given router
+    mntPt = require('../../../webgme-setup.json').components.routers['<%= restRouterName %>'].mount,
```
pmeijer: I like this approach.

pmeijer: (It's clear what is going on)

PR #1167, opened 2016-09-19
Share models

kecso:
The feature allows export and import of selected models.
The selection can be made in the ObjectBrowser or in the ModelEditor and the context menus provide the option for exporting.

Constraints:
- The root of the project cannot be exported this way
- Library elements cannot be exported this way (as the read-only property could be harmed)
- Only the minimal closure will be created upon export, so every element of the selection has to have a Meta base or the base also have to be in the selection
- Relations that going out from the selection will be lost
- In the place of import, the Meta should be identical (exact GUID match is necessary for the Meta bases) though it can have extra elements
- There will be no Meta rule check during import, so the resulting project could have Meta violations


src/client/js/Dialogs/ImportModel/ImportModelDialog.js
```diff
@@ -0,0 +1,121 @@
+/**
+ * @author kecso / https://github.com/kecso
+ */
+/*globals define, $*/
+/*jshint browser: true*/
```
pmeijer: Put these at the top of the file.

src/client/js/Panels/ModelEditor/ModelEditorControl.DiagramDesignerWidgetEventHandlers.js
```diff
@@ -1207,6 +1207,12 @@ define(['js/logger',
             MENU_META_RULES_NODE = 'metaRulesNode',
             MENU_META_RULES_MODEL = 'metaRulesModel',
             MENU_EDIT_REPLACEABLE = 'editReplaceable',
+            MENU_EXPORT_MODELS = 'exportModels',
+            MENU_IMPORT_MODELS = 'importModels',
+            node,
+            i,
+            paths,
+            no_library_content_selected = true,
```
pmeijer: Avoid negation and fix the variable name styling

src/client/js/Panels/ModelEditor/ModelEditorControl.DiagramDesignerWidgetEventHandlers.js
```diff
@@ -1277,6 +1312,15 @@ define(['js/logger',
                     });
                 } else if (key === MENU_EDIT_REPLACEABLE) {
                     self._replaceBaseDialog(self._ComponentID2GmeID[selectedIds[0]]);
+                } else if (key === MENU_EXPORT_MODELS) {
+                    paths = [];
+                    for (i = 0; i < selectedIds.length; i += 1) {
+                        paths.push(self._ComponentID2GmeID[selectedIds[i]]);
+                    }
+                    exporters.exportModels(self._client, self._logger, paths);
```
pmeijer: add line-break after }

src/client/js/Utils/Exporters.js
```diff
@@ -0,0 +1,82 @@
...
+                }
+
+                if (typeof callback === 'function') {
+                    callback(err);
+                }
+            }
+        );
+    }
+
+    function exportModels(client, logger, selectedIds, callback) {
```
pmeijer: Why not have an option withAssets here?

src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
```diff
@@ -423,14 +424,14 @@ define(['js/logger',
                         assetless: {
                             name: 'with assets',
                             callback: function (/*key, options*/) {
-                                self._exportProject(true);
+                                exporters.exportProject(self._client, self._logger, null, false);
```
pmeijer: withAssets parameter should be true

src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
```diff
@@ -423,14 +424,14 @@ define(['js/logger',
...
                             },
                             icon: false
                         },
                         assetfull: {
                             name: 'without assets',
                             callback: function (/*key, options*/) {
-                                self._exportProject(false);
+                                exporters.exportProject(self._client, self._logger, null, true);
```
pmeijer: withAssets parameter should be false

src/client/js/Utils/SaveToDisk.js
```diff
@@ -54,10 +54,79 @@ define(['blob/BlobClient'], function (BlobClient) {
         });
     }
 
+    function exportProject(client, logger, projectParams, withAssets, callback) {
```
pmeijer: This should be removed, right? It's in Utils/Exporters

src/common/core/librarycore.js
```diff
@@ -216,6 +216,288 @@ define([
                 self.setProperty(overlay, to, fromOverlay);
             }
 
+            function isPathInSubTree(fullPath, subTreePath) {
```
pmeijer: Couldn't this simply check:
`fullPath.length >= subTreePath.length && fullPath.indexOf(subTreePath) === 0`

kecso: no:
full = '/a/be/dee'
subtree = '/a/b'

it would make the simple check true, however this is not a container of the full path...

src/client/js/Utils/Exporters.js
kecso: It seemed too overcrowded to have multiple options (in the Model editor's context menu) for exporting models, but we might add the option on the ObjectBrowser and have the 'default' true in the model-editor

pmeijer: Agree

src/common/core/librarycore.js
pmeijer: You're right, and yeah I think that would be more efficient.

pmeijer:
- Change exported file extension to `.webgmexm` and check in the import dialog (still keep the checks done on the server).
- In the exported file-name only keep the first 6 digits in the commitHash.
- `selectionInfo.relations` and `selectionInfo.losses` -> `selectionInfo.relations: {preserved: {..}, lost: {..}}`

pmeijer:
Found a bug, the import doesn't give an error when one of the meta-bases doesn't exist. It makes a commit but no apparent changes are seen.

PR #1168, opened 2016-09-19
Restrict the usage of pointers named member

pmeijer:
Fixes:
- Restricts users from naming pointers 'member'
- Allow the look-up of sets named 'member'
- Fixes bug in meta-editor that removed/edit all set relationships between two meta-nodes.
- Improve the look-up in validPointers/Sets-getters in metacore

### DIFF
--- a/src/common/core/metacore.js
+++ b/src/common/core/metacore.js
@@ -222,12 +222,14 @@ define([
         };
 
         this.getValidPointerNames = function (node) {
-            var validNames = self.getPointerNames(getMetaNode(node)) || [],
+            var metaDefNode = getMetaNode(node),
+                validNames = self.getPointerNames(metaDefNode) || [],
                 i,
                 validPointerNames = [],
                 metaPointerNode, max;
+
             for (i = 0; i < validNames.length; i++) {
-                metaPointerNode = getMetaPointerNode(node, validNames[i]);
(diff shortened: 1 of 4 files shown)


src/common/core/metacore.js
```diff
@@ -222,12 +222,14 @@ define([
...
-            var validNames = self.getPointerNames(getMetaNode(node)) || [],
+            var metaDefNode = getMetaNode(node),
+                validNames = self.getPointerNames(metaDefNode) || [],
                 i,
                 validPointerNames = [],
                 metaPointerNode, max;
+
             for (i = 0; i < validNames.length; i++) {
-                metaPointerNode = getMetaPointerNode(node, validNames[i]);
+                metaPointerNode = self.getChild(metaDefNode, CONSTANTS.META_POINTER_PREFIX + validNames[i]);
```
kecso: Why replacing the getMetaPointerNode in some places and leave untouched in others?
Either the functions should be changed or it should be removed from everywhere.
(or just I am not seeing the distinction among these usages)

pmeijer: For the cases I replaced it we already have the `metaDefNode`, plus we are iterating over the `validNames`. The later part means that the check and the additional retrieval of `validNames` in the getMetaPointerNode is redundant.
This is not the case for a places where it is used.

PR #1170, opened 2016-09-20
Update from file

kecso:
With the import command line, a branch of a project can be updated from a file.

### DIFF
--- a/test/bin/import.spec.js
+++ b/test/bin/import.spec.js
@@ -97,7 +97,7 @@ describe('import CLI tests', function () {
         projectName = null;
         projectId = null;
 
-        importCLI.should.have.property('main');
+        expect(importCLI).to.have.property('main');
     });
 
     it('should fail if mandatory parameters missing', function (done) {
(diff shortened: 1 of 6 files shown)


test/bin/import.spec.js
```diff
@@ -15,7 +15,6 @@ describe('import CLI tests', function () {
         importPath = './test/bin/import/project.webgmex',
         projectName,
         projectId,
-        __should = testFixture.should,
```
pmeijer: If you run this test file on its own, line 100 will fail. Since that's the only should, you can just change it to expect..

test/seeds/seeds.spec.js
```diff
@@ -81,6 +81,7 @@ describe('Seeds', function () {
         var agent = superagent.agent();
 
         agent.get(serverBaseUrl + '/api/seeds', function (err, res) {
+            console.error(res.body);
```
pmeijer: console...

PR #1180, opened 2016-09-23
Fixes for #1177 relid collisions from base class

pmeijer:
The property here means the newly introduced `MINIMAL_RELID_LENGTH_PROPERTY`

- The property should be propagated at setBase.
 - When checking the length of the children at the node getting a new base it neglects relids lengths > `MAXIMUM_STARTING_RELID_LENGTH`
- The property should be propagated recursively.
- The property should not be copied to copies of a node.

### DIFF
--- a/src/common/core/coretype.js
+++ b/src/common/core/coretype.js
@@ -688,11 +688,15 @@ define([
             ASSERT(self.isValidNewParent(parent, node),
                 'New parent would create loop in containment/inheritance tree.');
             var base = node.base,
+                minRelidLength = innerCore.getProperty(parent, CONSTANTS.MINIMAL_RELID_LENGTH_PROPERTY) || 0,
+                takenRelids = self.getChildrenRelids(parent, true),
                 moved;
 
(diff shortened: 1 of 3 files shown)


src/common/core/coretype.js
```diff
@@ -1102,16 +1144,7 @@ define([
         };
 
         this.processRelidReservation = function (node, relid) {
```
kecso: why not have only the 'length' function?

test/common/core/coretype.spec.js
```diff
@@ -6,7 +6,7 @@
  */
 var testFixture = require('../../_globals.js');
 
-describe('coretype', function () {
+describe.only('coretype', function () {
```
kecso: .only

src/common/core/coretype.js
pmeijer: I kept the one defined which is easier to use for most cases (a new relid was generated - process the bases..). The one taking a number is really only used in `setBase`.

PR #1206, opened 2016-10-13
Closes #1205 by adding optional server.handle to be passed from config.

### LINKED ISSUE #1205: Enable webGME to run under systemd control
The obvious approach is to add the following to <gme-app>/config/config.default.js
```
...
var systemdSocket = require("systemd-socket");
...
config.server.port = systemdSocket() || 3000;
...
```
The difficulty is that webgme/config/validator.js has an overly strict constraint, i.e. more restrictive than the [method it is protecting](https://nodejs.org/api/http.html#http_server_listen_handle_callback).
Which is [called from here]( https://github.com/webgme/webgme/blob/master/src/server/standalone.js#L216).
At issue is the fact that systemdSocket(0) does not return a number but rather an object of the form '{fd: 3}'.
The change would be in https://github.com/webgme/webgme/blob/master/config/validator.js#L194
Relaxing the constraint to allow an object as well as a number.

pmeijer:


phreed:
Looks like it works.
Thanks

PR #1209, opened 2016-10-14
Remove reverse overlay

kecso:
This function removes the reverse relations on data level. From now on, whenever we need the inverse relations, we will compute it based on the forward ones (it was always the same node that held the information about both direction of a relation).
To ease the computation burden we also introduce the following functionalities:
- the core will hold a cache for the inverse-relation-collections that is based on the hash of the node object
- if a node changes and it already has a computed inverse-collection, that will also change with it (and mutate if necessary, just the regular data would do)

As a side-effect, the patch objects that are being sent with every commit event, got smaller, but we had to add some information about the destinations of the changing relations, so the UI event generation would remain possible.


src/common/core/corerel.js
```diff
@@ -171,17 +137,73 @@ define([
...
+            storeNewInverseOverlays(self.getRoot(node));
+
+            return persisted;
+        };
         //</editor-fold>
 
         //<editor-fold=Added Methods>
+        this.getInverseOverlayOfNode = function (node) {
+            var hash = self.getHash(node),
+                inverseOverlays = {},
```
pmeijer: It's a little misleading to initialize the inverseOverlays up here.

src/common/core/corerel.js
```diff
@@ -171,17 +137,73 @@ define([
...
             //TODO this is needed as now we work with modified data as well
             if (name === CONSTANTS.MUTABLE_PROPERTY) {
                 return false;
             }
-            return name.slice(-CONSTANTS.COLLECTION_NAME_SUFFIX.length) !==
-                   CONSTANTS.COLLECTION_NAME_SUFFIX;
+            // return name.slice(-CONSTANTS.COLLECTION_NAME_SUFFIX.length) !==
+            //     CONSTANTS.COLLECTION_NAME_SUFFIX;
+
+            return true;
```
pmeijer: Why did this change? Or it should've changed anyhow?

src/common/core/corerel.js
```diff
@@ -707,88 +830,72 @@ define([
...
-                        if (!self.isPointerName(name) && name !== CONSTANTS.MUTABLE_PROPERTY) {
-                            name = name.slice(0, -CONSTANTS.COLLECTION_NAME_SUFFIX.length);
-                            if (self.isPointerName(name) && names.indexOf(name) < 0) {
-                                names.push(name);
-                            }
+                inverseOverlays = self.getInverseOverlayOfNode(node);
+                if (inverseOverlays[target]) {
+                    for (name in inverseOverlays[target]) {
+                        if (names.indexOf(name) === -1) {
+                            names.push(name);
```
pmeijer: Why does the logic change here (no more isPointerName.. etc.)?

src/common/storage/dataPatches.js
```diff
@@ -0,0 +1,64 @@
+/*globals define*/
```
pmeijer: This should be named `dataPatchers`

pmeijer: Or maybe even `dataConverters`

src/common/storage/util.js
```diff
@@ -149,6 +150,24 @@ define([
...
+     * nothing will happen to the object.
+     *
+     * @param {Object} dataObject
+     */
+    function patchDataObject(dataObject) {
+        var dataVersion = dataObject.__v || '0.0.0',
+            myVersion = CONSTANTS.VERSION || '0.0.0';
+
+        if (dataPatches[dataVersion] && typeof dataPatches[dataVersion][myVersion] === 'function') {
+            dataPatches[dataVersion][myVersion](dataObject);
```
pmeijer: This is why it should be renamed. Looking here it seems that `dataPatches` are some patch objects for the data...

src/common/util/jsonPatcher.js
```diff
@@ -273,7 +328,7 @@ define([
 
     function _getChangedNodesRec(patch, res, hash, gmePath) {
         var nodePatches = patch[hash] && patch[hash].patch, // Changes regarding node with hash
-            i,
+            i, j,
```
pmeijer: Maybe we could generate all the changed node data at patch instead?
A `//TODO:...` would do for now.

test-karma/client/js/client.spec.js
```diff
@@ -1976,7 +1976,7 @@ describe('GME client', function () {
                     expectedEventNodes.push('');
                     // 2. the base of the copied node
                     expectedEventNodes.push('/701504349');
-                    // 3,4. the two memebers of the copied node
+                    // 3,4. the two members of the copied node
```
pmeijer: For some reason I tend to write memebers when I mean members.

config/config.default.js
```diff
@@ -68,6 +68,7 @@ var path = require('path'),
 
         core: {
             enableCustomConstraints: false,
+            inverseRelationsCacheSize: 10000
```
pmeijer: I would put this to the same value as the regular cache (2000).

src/common/core/corerel.js
```diff
@@ -171,17 +137,73 @@ define([
...
+
+            storeNewInverseOverlays(self.getRoot(node));
+
+            return persisted;
+        };
         //</editor-fold>
 
         //<editor-fold=Added Methods>
+        this.getInverseOverlayOfNode = function (node) {
+            var hash = self.getHash(node),
```
kecso: It should be more of a look-up.
The normalization in most of the cases should not be computationally heavy, but will remove after the first branch just in case.

src/common/core/corerel.js
kecso: After this change, the 'ptr-inv' can be a regular pointer name. That is why I have changed the function. Though we are not yet propagated this to the UI, so we can keep the check as it was.

src/common/core/corerel.js
kecso: The inverseOverlay should only contain names that are allowed pointer names, so there is no point in running the check.

src/common/util/jsonPatcher.js
kecso: I am not quite sure what you mean by this comment.
If you think of generating an 'event' field for the patch object, then I believe it should be possible (though we still have to go through all the nodes in our territory to gather the events, but at least we could save the checking of every operation in a patch object...)

PR #1211, opened 2016-10-17
Do not use raphael api to get length and bboxes for connections.

pmeijer:
Opening up Will 5k nodes in one parent model goes down from about 75 to 23 seconds.

### DIFF
--- a/src/client/js/Widgets/DiagramDesigner/Connection.js
+++ b/src/client/js/Widgets/DiagramDesigner/Connection.js
@@ -500,6 +500,36 @@ define([
         }
     };
 
+    Connection.prototype._calcRawBoundingBox = function () {
+        var result = {
+                x: 0,
+                y: 0,
+                x2: 0,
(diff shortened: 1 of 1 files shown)


src/client/js/Widgets/DiagramDesigner/Connection.js
```diff
@@ -1612,45 +1630,48 @@ define([
...
+        var self = this,
+            TEXT_OFFSET = 5;
 
         this._hideTexts();
 
-        this.skinParts.textContainer = this._textContainer.clone();
-        this.skinParts.textContainer.attr('id', TEXT_ID_PREFIX + this.id);
+        function drawName() {
+            var totalLength = self.calcTotalLength(),
+                pathCenter = self.skinParts.path.getPointAtLength(totalLength / 2);
```
kecso: shouldn't the getPointAtLength be revised as well?

PR #1212, opened 2016-10-18
Collect instances

kecso:
With the new functionality it will be possible to collect instances paths even if we are approaching instances that are inherited children. Also the loadInstances shortcut to access instances could speed up some future functions.

### DIFF
--- a/src/common/core/core.js
+++ b/src/common/core/core.js
@@ -2019,7 +2019,7 @@ define([
          *
          * @func
          */
-        this.getInstancesPaths = core.getInstancesPaths;
+        this.getInstancePaths = core.getInstancePaths;
 
         /**
          * Loads all the instances of the given node.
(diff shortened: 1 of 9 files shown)


pmeijer:
Looks good, just awaiting some native-speaking feed-back regarding the `getInstancesPaths` vs `getInstancePaths`

pmeijer:
Both @brollb and @phreed preferred `getInstancePaths`

src/common/core/core.js
```diff
@@ -2010,6 +2010,25 @@ define([
...
+
+        /**
+         * Collects the paths of all the instances of the given node.
+         * @param {module:Core~Node} node - the node in question.
+         *
+         *@return {string[]} The function returns an array of the absolute paths of the instances.
+         *
+         * @func
+         */
+        this.getInstancesPaths = core.getInstancesPaths;
```
pmeijer: getInstancePaths

PR #1213, opened 2016-10-18
Allow for upload of large blob files and give progress feed-back on UI.

### LINKED ISSUE #1145: Large file (asset) upload for plugins fails
Can't attach files w/ size greater than or equal to 1.5gb

pmeijer:
This closes #1145.

To enable larger file-uploads increase the config.server.timeout which sets the [timeout for the http-server ](https://nodejs.org/api/http.html#http_server_timeout).

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -168,7 +168,7 @@ To configure the default behaviour of individual components (e.g. plugins, ui-wi
  - Port the server is hosted from.
 - `config.server.handle = null`
  - Optional handle object passed to [server.listen](https://nodejs.org/api/http.html#http_server_listen_handle_callback) (aligning port must still be given).
-- `config.server.timeout = 0`
+- `config.server.timeout = -1`
  - If greater than -1 will set the [timeout property of the http-server](https://nodejs.org/api/http.html#http_server_timeout). (This can be used to enable large, > 1Gb, file uploads.)
 - `config.server.maxWorkers = 10`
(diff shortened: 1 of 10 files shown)


src/server/standalone.js
```diff
@@ -148,6 +148,10 @@ function StandAloneServer(gmeConfig) {
 
         __httpServer = Http.createServer(__app);
 
+        if (gmeConfig.server.timeout > 0) {
+            __httpServer.timeout = gmeConfig.server.timeout;
```
brollb: What if this is 0? Shouldn't we still set it? According to the http docs:

>Set to 0 to disable any kind of automatic timeout behavior on incoming connections.

https://nodejs.org/api/http.html#http_server_timeout

config/README.md
```diff
@@ -168,6 +168,8 @@ To configure the default behaviour of individual components (e.g. plugins, ui-wi
  - Port the server is hosted from.
 - `config.server.handle = null`
  - Optional handle object passed to [server.listen](https://nodejs.org/api/http.html#http_server_listen_handle_callback) (aligning port must still be given).
+- `config.server.timeout = 0`
```
pmeijer: It's not being disable since it's not > 0. However we should allow disabling too.

src/server/standalone.js
pmeijer: Good point, which should allow for that too.

PR #1217, opened 2016-10-21
Use new client API functions

pmeijer:
This gets rid of the warnings from the deprecated function calls.


brollb:
How long are the old functions going to be supported?

pmeijer:
At least till we bump the version to 3 (which isn't planned yet).

pmeijer:
Cherry pick this to v2.6.x as well

Issue #1063, opened 2016-07-16
incorrect active object change event


brollb:
After loading the page, pressing the back arrow will result in the "active object changed" event being fired although the active object has not changed.

PR #1223, opened 2016-10-27
Closes #1222, #1063 and various fixes for Set- and CrossCut-editor.

### LINKED ISSUE #1222: Starting the UI from SetEditor throws exception.
And when not it renders the objects on the canvas twice.

pmeijer:
Fixes a bug introduced by #1217 that disabled adding new members to CrossCuts.

In addition connections are always rendered as boxes in the set-editor.

In CrossCuts - the autorouter option is disabled (until it supports connections to connections).


pmeijer:
This should be cherry-picked to master as well.

Issue #1228, opened 2016-11-03
Getting setMemberAttributes/Registries excludes properties that shouldn't be excluded.


pmeijer:
The predicate when getting the properties (attribute/registry names in this case) checks for isValidRelid, it should exclude properties starting with an underscore.

pmeijer:
Closed by #1229

Issue #1227, opened 2016-11-02
Add core methods for accessing set registries and attributes.


pmeijer:
`getSetRegistryNames`
`getOwnSetRegistryNames`
`getSetRegistry`
`getOwnSetRegistry`
`setSetRegistry`
`delSetRegistry`

and

`getSetAttributeNames`
`getOwnSetAttributeNames`
`getSetAttribute`
`getOwnSetAttribute`
`setSetAttribute`
`delSetAttribute`

In addition the following methods should be added for consistency;
`getOwnSetNames`
`getMemberOwnRegistry`
`getMemberOwnAttribute`

pmeijer:
Closed by #1229

PR #1229, opened 2016-11-07
Add support for documentation in MetaEditor, fixes #903 and also #1227 and #1228

### LINKED ISSUE #903: Provide Document Nodes in META
Provide some type of document node for the MetaSheets.
Having document nodes in the composition visualizer are great.
I have no idea how this might be implemented but it would be very helpful.

pmeijer:
See wiki for details
https://github.com/webgme/webgme/wiki/Documenting-Models

This also adds to #1232 w.r.t. creating/moving nodes inside an inherited child.

### DIFF
--- a/src/client/js/Panels/MetaEditor/MetaDocItem.js
+++ b/src/client/js/Panels/MetaEditor/MetaDocItem.js
@@ -3,8 +3,12 @@
  * @author pmeijer / https://github.com/pmeijer
  */
 
-define(['./MetaEditorConstants'], function (META_EDITOR_CONSTANTS) {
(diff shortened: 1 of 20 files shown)


src/client/js/Panels/MetaEditor/MetaDocItem.js
```diff
@@ -0,0 +1,142 @@
...
+                                getRegistry: function (name) {
+                                    return self[name];
+                                }
+                            };
+                        },
+                        setAttribute: function (_id, name, value) {
+                            if (name === 'documentation' || name === 'name') {
+                                self.setProperty(name, value);
+                            } else {
+                                console.warn('unexpected attribute in MetaDocItem', name, value);
```
kecso: couldn't it use the logger of the client (or even better fork it) somehow?

PR #1231, opened 2016-11-08
Fixes #892 add better feedback and better handling of invalidly named pointers/sets

### LINKED ISSUE #892: Naming a pointer 'base' is not allowed
Naming a pointer 'base' is not allowed.
There is no indication that 'base' is a reserved word.
Nor is there a compelling reason that it should be reserved.

pmeijer:
Now it checks for all meta-definitions to avoid conflicts between set and pointer names.
It also ensures that there are no illegal characters or suffixes. 

With violations it also gives a descriptive message.

### DIFF
--- a/src/client/js/Panels/MetaEditor/MetaEditorControl.js
+++ b/src/client/js/Panels/MetaEditor/MetaEditorControl.js
@@ -1240,46 +1240,71 @@ define(['js/logger',
     };
 
     MetaEditorControl.prototype._createPointerRelationship = function (sourceID, targetID, isSet) {
-        var sourceNode = this._client.getNode(sourceID),
-            targetNode = this._client.getNode(targetID),
+        var client = this._client,
+            sourceNode = client.getNode(sourceID),
+            targetNode = client.getNode(targetID),
+            metaNodes = client.getAllMetaNodes(true),
             pointerMetaDescriptor,
+            existingNames,
             existingPointerNames,
-            notAllowedPointerNames,
-            self = this;
+            notAllowedPointerNames;
+
+        function getExistingNames() {
+            var nodeObj = sourceNode,
+                lastMetaNode,
+                i,
+                instancePaths;
+
+            // Get the meta-node "last" in the inheritance chain
+            do {
+                instancePaths = nodeObj.getInstancePaths();
+
+                lastMetaNode = null;
+                for (i = 0; i < instancePaths.length; i += 1) {
+                    if (metaNodes[instancePaths[i]]) {
+                        lastMetaNode = metaNodes[instancePaths[i]];
+                        break;
+                    }
+                }
+
+                if (lastMetaNode) {
+                    nodeObj = lastMetaNode;
+                }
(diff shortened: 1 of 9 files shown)


src/client/js/Panels/MetaEditor/MetaEditorControl.js
```diff
@@ -1174,46 +1174,71 @@ define(['js/logger',
...
+            targetNode = client.getNode(targetID),
+            metaNodes = client.getAllMetaNodes(true),
             pointerMetaDescriptor,
+            existingNames,
             existingPointerNames,
-            notAllowedPointerNames,
-            self = this;
+            notAllowedPointerNames;
+
+        function getExistingNames() {
```
kecso: how would this function deal with the situation when typeA has two instances typeAone and typeAtwo that both introduces new names?
why the existing names do not cover all existing names throughout the Meta (or I read it wrong?)?

src/client/js/Panels/MetaEditor/MetaEditorControl.js
```diff
@@ -1174,46 +1174,71 @@ define(['js/logger',
...
-            notAllowedPointerNames = _.union(sourceNode.getSetNames(), sourceNode.getPointerNames());
+            existingNames = getExistingNames();
+
             if (isSet === true) {
-                //this is a pointer list
-                existingPointerNames = _.difference(sourceNode.getSetNames() || [],
-                    sourceNode.getValidAspectNames());
+                //this is a set
+                existingPointerNames = existingNames.sets;
+                notAllowedPointerNames = _.union(existingNames.pointers, existingNames.aspects);
```
kecso: yeah,  I think in theory aspects could use the same names, but it might be that on some level they share the same ids inside the node...

src/client/js/Widgets/MetaEditor/MetaEditorPointerNamesDialog.js
```diff
@@ -67,10 +62,51 @@ define([
...
+                result.hasViolation = true;
+                result.message = (isSet ? 'Set' : 'Pointer') + ' must have a non-empty name.';
+            } else if (notAllowedPointerNames.indexOf(name) > -1) {
+                result.hasViolation = true;
+                result.message = 'Name "' + name + '" is already used for a ' + (isSet ? 'pointer' : 'set') + ' or ' +
+                'an aspect.';
+            } else if (REGEXP.DOCUMENT_KEY.test(name) === false) {
+                result.hasViolation = true;
+                result.message = 'Name "' + name + '" contains illegal characters, it may not contain "." ' +
+                    'or start with "$".';
```
kecso: aren't we forbid using '_' as starting character as well?

src/client/js/Panels/MetaEditor/MetaEditorControl.js
pmeijer: I changed this, now it only checks what's defined for the current node. If there are collisions down the inheritance chains - those issues are caught by the MetaEditor using #1258

PR #1232, opened 2016-11-09
Process the min_relid_length of parent of when inherited child modified.

pmeijer:
The potential issue here was that if we have a node `P` with a child `C` and an instance `P'` with the inherited child `C'`.
The following could lead to corrupt models:
1) we set an attribute of `C'`.
2) we delete `C`.
3) without reloading `C` create a new child that gets the same relid as `C` (and `C'`).
4) Now if we load `C'` it would have the attribute set at 1).

This applies to setting pointers, registries and adding sets, set-members and set-memberAttributes/Registries to the inherited child.

This PR also gets rid of generation of short relids for children (member entries) inside sets. These are set to have length 6 (MAXIMUM_STARTING_RELID_LENGTH + 1). Relid collisions are still checked among own entries and inherited ones. However since the relid is greater than MAXIMUM_STARTING_RELID_LENGTH, the propagation of MINIMAL_RELID_LENGTH_PROPERTY is not made. (This is important because otherwise the FCO would have this set for every single set defined in the model.)

### DIFF
--- a/src/common/core/coretype.js
+++ b/src/common/core/coretype.js
@@ -456,8 +456,13 @@ define([
         function processNewRelidLength(node, newMinLength) {
             var currMinLength;
 
-            newMinLength = newMinLength > CONSTANTS.MAXIMUM_STARTING_RELID_LENGTH ?
-                CONSTANTS.MAXIMUM_STARTING_RELID_LENGTH : newMinLength;
+            if (newMinLength > CONSTANTS.MAXIMUM_STARTING_RELID_LENGTH + 1) {
+                logger.debug('Minimum relid length surpassed threshold, not propagating at all', newMinLength);
+                return;
+            } else if (newMinLength > CONSTANTS.MAXIMUM_STARTING_RELID_LENGTH) {
+                newMinLength = CONSTANTS.MAXIMUM_STARTING_RELID_LENGTH;
+                logger.debug('Minimum relid length reached threshold, only propagating threshold', newMinLength);
+            }
 
             node = node.base;
             while (node) {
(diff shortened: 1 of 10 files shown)


src/common/core/coretype.js
```diff
@@ -604,13 +604,19 @@ define([
         };
 
         this.setPointer = function (node, name, target) {
+            var parent;
             innerCore.setPointer(node, name, target);
+
             if (isInheritedChild(node)) {
-                this.setProperty(node, CONSTANTS.INHERITED_CHILD_HAS_OWN_RELATION_PROPERTY, true);
+                self.setProperty(node, CONSTANTS.INHERITED_CHILD_HAS_OWN_RELATION_PROPERTY, true);
+                parent = self.getParent(node);
```
kecso: why this is in a separate command when we do not check anything for the parent?
and what if the parent is null? (probably it is enough to handle inside the function, but somewhere we should deal with it)

src/common/core/coretype.js
```diff
@@ -604,13 +604,19 @@ define([
...
+                self.processRelidReservation(parent, self.getRelid(node));
             }
 
             if (isInheritedChild(target)) {
-                this.setProperty(target, CONSTANTS.INHERITED_CHILD_HAS_OWN_RELATION_PROPERTY, true);
+                self.setProperty(target, CONSTANTS.INHERITED_CHILD_HAS_OWN_RELATION_PROPERTY, true);
+                parent = self.getParent(target);
```
kecso: same here

src/common/core/coretype.js
```diff
@@ -975,6 +981,35 @@ define([
...
+        this.setAttribute = function (node, name, value) {
+            var parent;
+
+            innerCore.setAttribute(node, name, value);
+
+            if (isInheritedChild(node)) {
+                // If the node is an inherited child - data is now stored for it.
+                // And to prevent application of this data after deletion and creation of
+                // new nodes at the base - we process the relid.
+                parent = self.getParent(node);
```
kecso: and here

src/common/core/coretype.js
```diff
@@ -975,6 +981,35 @@ define([
...
+        this.setRegistry = function (node, name, value) {
+            var parent;
+
+            innerCore.setRegistry(node, name, value);
+
+            if (isInheritedChild(node)) {
+                // If the node is an inherited child - data is now stored for it.
+                // And to prevent application of this data after deletion and creation of
+                // new nodes at the base - we process the relid.
+                parent = self.getParent(node);
```
kecso: and here

pmeijer:
This should be cherry-picked into 2.6.x

PR #1236, opened 2016-11-14
Data conversion should not modify objects on storage level.

pmeijer:
Instead when the data is loaded for a core-node a new data is created matching the new version.
This PR also:
- Bumps the object version to `1.1.0` (from `1.0.0`).
- By default ensures that the patched object on the server matches the given hash.

Related to #1209


pmeijer:
This should be cherry-picked into v2.6.x

PR #1238, opened 2016-11-15
Line preferences

kecso:
Now the preferences tab in the property editor will have the following additional values:
image

They were available for edit already, but with presenting them on the property editor the inheritance rules become visible and they became more transparent.
The values are visible for any node, even if they not visualized as connection (as they might be a visualizer that leverages these registry values).


src/client/js/Controls/PropertyGrid/Widgets/styles/SvgSelectWidget.css
```diff
@@ -0,0 +1,30 @@
+/**
```
pmeijer: No scss?

src/client/assets/line/lineSvgs.js
```diff
@@ -0,0 +1,440 @@
...
+        '<text x="43" y="9" text-anchor="middle" font="10px &quot;Arial&quot;" stroke="none" fill="#000000" ' +
+        'font-size="8px" style="-webkit-tap-highlight-color: rgba(0, 0, 0, 0); text-anchor: middle; ' +
+        'font-style: normal; font-variant: normal; font-weight: normal; font-stretch: normal; font-size: 8px; ' +
+        'line-height: normal; font-family: Arial;"><tspan dy="9" ' +
+        'style="-webkit-tap-highlight-color: rgba(0, 0, 0, 0);">$</tspan></text></svg>';
+
+    return line_svg_directory;
+});
+
+/*
```
pmeijer: Any reason this is kept?

kecso: have not added the bezier options yet so I figured that instead of debugging again we can keep it for a while, but I will attach a TODO if that is fine

src/client/js/Controls/PropertyGrid/Widgets/styles/SvgSelectWidget.css
kecso: still having hard time with the scss, but as soon as they will be forced I will only use those.

PR #1249, opened 2016-11-23
blob-fs-clean-up script

kecso:
As the blobStorage is a write-once storage, there is a need for having a utility that could clean the unused ones.

If a single blobStorage connected to multiple WebGME servers, then the user had to manually merge the unused lists to really only clear those that do not have references in any server.

- the commit should be cherry-picked to 2.7.x if we want to provide the functionality as early as possible -


src/bin/blob_fs_clean_up.js
```diff
@@ -0,0 +1,355 @@
...
+        usedHashesObject,
+        i;
+
+    blobClient.listObjects('wg-metadata')
+        .then(function (all) {
+            allHashesArray = all;
+            return getUsedMetaHashes(metadatastorage, storage, blobClient);
+        })
+        .then(function (used) {
+            usedHashesObject = used;
```
pmeijer: usedHashesObject seems redundant

src/bin/blob_fs_clean_up.js
```diff
@@ -0,0 +1,355 @@
...
+            } else if (metaData.contentType === CONTENT_TYPES.COMPLEX) {
+                for (complexEntry in metaData.content) {
+                    if (metaData.content[complexEntry].contentType === CONTENT_TYPES.SOFT_LINK) {
+                        softLinks.push(metaData.content[complexEntry].content);
+                    }
+                }
+            }
+            deferred.resolve(softLinks);
+        })
+        .catch(deferred.reject);
```
pmeijer: You could expect to see some errors here. The only check for the assets is if it is a blob-hash.

src/bin/blob_fs_clean_up.js
```diff
@@ -0,0 +1,355 @@
...
+}
+
+function gatherSoftLinks(blobClient, metaHash) {
+    var deferred = Q.defer(),
+        softLinks = [],
+        complexEntry;
+
+    blobClient.getMetadata(metaHash)
+        .then(function (metaData) {
+            if (metaData.contentType === CONTENT_TYPES.SOFT_LINK) {
```
pmeijer: Is this ever true?

PR #1251, opened 2016-11-26
Adjusted client.copyNode to return nodeId. Fixes #1250

### LINKED ISSUE #1250: client.copyNode should return new node id

brollb:


pmeijer:
Thanks Brian

PR #1260, opened 2016-12-16
Better error handling in core

kecso:
This enhancement introduces extra checks at the upper layers of Core and unifies its behavior regarding errors. Whenever an ASSERTION fails or some check shows invalid inputs, the Core will throw one of the following errors:
- CoreIllegalArgumentError - if the type of the input parameters is not what it should be
- CoreIllegalOperationError - if the set of input parameters are correct but the request or the operation do not apply to the current context. Here we followed the basic javascript principles in terms that whenever the user try to access a 'field' of a 'field' that does not exist, we throw. For example if someone tries to get the member attributes of an non-existing member. Trying to modify read-only nodes are captured within this category.
- CoreAssertError - happens if some internal ASSERTION fails, it triggers some fault inside the core and should typically be checked by the developer team, not the one who uses it.

The throwing behavior is followed in the client as well, meaning the from this enhancement the UI developers should be prepared for exceptions.

It also updates the function descriptions of the core (mentioning the promise like behavior).

### DIFF
--- a/src/common/core/CoreAssertError.js
+++ b/src/common/core/CoreAssertError.js
@@ -12,6 +12,8 @@ define([], function () {
         error.name = this.name = 'CoreAssertError';
         this.message = error.message;
         this.stack = error.stack;
+
+        return error;
     }
 
     CoreAssertError.prototype = Object.create(Error.prototype);
--- a/src/common/core/CoreAssertError.js
+++ b/src/common/core/CoreAssertError.js
@@ -1,0 +1,21 @@
+/*globals define*/
+/*jshint node: true, browser: true*/
+
+/**
+ * @author kecso / https://github.com/kecso
+ */
+
+define([], function () {
+    'use strict';
+    function CoreAssertError() {
+        var error = Error.apply(this, arguments);
+        error.name = this.name = 'CoreAssertError';
+        this.message = error.message;
+        this.stack = error.stack;
+    }
+
+    CoreAssertError.prototype = Object.create(Error.prototype);
+    CoreAssertError.prototype.constructor = CoreAssertError;
+
+    return CoreAssertError;
+});
--- a/src/common/core/core.js
+++ b/src/common/core/core.js
@@ -771,11 +771,15 @@ define([
         this.createNode = function (parameters) {
             if (parameters) {
                 ensureType(parameters, 'parameters', 'object');
-                ensureNode(parameters.parent, 'parameters.parent');
-                if (parameters.base) {
+                if (parameters.hasOwnProperty('parent') &&
+                    parameters.parent !== null && parameters.parent !== undefined) {
+                    ensureNode(parameters.parent, 'parameters.parent');
+                }
+                if (parameters.hasOwnProperty('base') &&
+                    parameters.base !== null && parameters.base !== undefined) {
                     ensureNode(parameters.base, 'parameters.base');
                 }
-                if (parameters.guid) {
+                if (parameters.hasOwnProperty('guid') && parameters.guid !== undefined) {
                     ensureGuid(parameters.guid, 'parameters.guid');
                 }
             }
@@ -790,10 +794,13 @@ define([
          * @throws {CoreIllegalOperationError} If the context of the operation is not allowed.
          * @throws {CoreAssertError} If some internal error took place inside the core layers.
          */
-        this.deleteNode = function (node) {
+        this.deleteNode = function (node, technical) {
             ensureNode(node, 'node');
+            if (technical !== null && technical !== undefined) {
+                ensureType(technical, 'technical', 'boolean');
+            }
 
-            return core.deleteNode(node);
+            return core.deleteNode(node, technical);
         };
 
         /**
@@ -1079,7 +1086,7 @@ define([
          * Sets the target of the pointer of the node.
          * @param {module:Core~Node} node - the node in question.
          * @param {string} name - the name of the pointer in question.
-         * @param {module:Core~Node} target - the new target of the pointer.
+         * @param {module:Core~Node|null} target - the new target of the pointer.
          *
          * @throws {CoreInputError} If some of the parameters doesn't match the input criteria.
          * @throws {CoreIllegalOperationError} If the context of the operation is not allowed.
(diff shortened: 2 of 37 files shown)


src/common/core/core.js
```diff
@@ -1272,60 +2211,92 @@ define([
...
+         * @throws {CoreIllegalOperationError} If the context of the operation is not allowed.
+         * @throws {CoreAssertError} If some internal error took place inside the core layers.
          */
-        this.delConstraint = core.delConstraint;
+        this.delConstraint = function (node, name) {
+            ensureNode(node, 'node');
+            ensureType(name, 'name', 'string');
+            var names = core.getConstraintNames(node);
+            if (names.indexOf(name) === -1) {
+                throw new CoreIllegalOperationError('Cannot remove unknown constraint.');
```
pmeijer: SHould not throw

src/common/core/core.js
```diff
@@ -1380,9 +2374,21 @@ define([
...
+         * @throws {CoreAssertError} If some internal error took place inside the core layers.
          */
-        this.isValidTargetOf = core.isValidTargetOf;
+        this.isValidTargetOf = function (node, source, name) {
+            ensureNode(node, 'node');
+            ensureNode(source, 'source');
+            ensureType(name, 'name', 'string');
+            var names = core.getValidPointerNames(source).concat(core.getValidSetNames(source));
+            if (names.indexOf(name) === -1) {
+                throw new CoreIllegalOperationError('Cannot get information about unknown pointer definition.');
```
pmeijer: I don't think this should throw

src/common/core/core.js
```diff
@@ -1542,31 +2600,50 @@ define([
...
+         * @throws {CoreIllegalOperationError} If the context of the operation is not allowed.
+         * @throws {CoreAssertError} If some internal error took place inside the core layers.
          */
-        this.delAttributeMeta = core.delAttributeMeta;
+        this.delAttributeMeta = function (node, name) {
+            ensureNode(node, 'node');
+            ensureType(name, 'name', 'string');
+            var names = core.getValidAttributeNames(node);
+            if (names.indexOf(name) === -1) {
+                throw new CoreIllegalOperationError('Cannot remove unknown attribute definition');
```
pmeijer: Should not throw

src/common/core/core.js
```diff
@@ -1651,24 +2745,38 @@ define([
...
+         * @throws {CoreIllegalOperationError} If the context of the operation is not allowed.
+         * @throws {CoreAssertError} If some internal error took place inside the core layers.
          */
-        this.delChildMeta = core.delChildMeta;
+        this.delChildMeta = function (node, path) {
+            ensureNode(node, 'node');
+            ensurePath(path, 'path');
+            var paths = core.getValidChildrenPaths(node);
+            if (paths.indexOf(path) === -1) {
+                throw new CoreIllegalOperationError('Cannot remove child definition of an unknown child.');
```
pmeijer: Should not trhwo

src/common/core/core.js
```diff
@@ -2026,58 +3323,66 @@ define([
...
+         * @throws {CoreIllegalOperationError} If the context of the operation is not allowed.
+         * @throws {CoreAssertError} If some internal error took place inside the core layers.
          */
-        this.delMixin = core.delMixin;
+        this.delMixin = function (node, path) {
+            ensureNode(node, 'node');
+            ensurePath(path, 'path');
+            var paths = core.getMixinPaths(node);
+            if (paths.indexOf(path) === -1) {
+                throw new CoreIllegalOperationError('Cannot remove unknown mixin.');
```
pmeijer: Should not throw

src/common/core/core.js
```diff
@@ -1766,47 +2918,77 @@ define([
...
-         *
-         * @func
+         * @throws {CoreInputError} If some of the parameters doesn't match the input criteria.
+         * @throws {CoreIllegalOperationError} If the context of the operation is not allowed.
+         * @throws {CoreAssertError} If some internal error took place inside the core layers.
          */
-        this.setAspectMetaTarget = core.setAspectMetaTarget;
+        this.setAspectMetaTarget = function (node, name, target) {
+            ensureNode(node, 'node');
+            ensureType(name, 'name', 'string');
```
pmeijer: This should be checked for 'ovr',  and starting with '_'

src/common/core/core.js
```diff
@@ -1725,32 +2858,51 @@ define([
...
+         * @throws {CoreAssertError} If some internal error took place inside the core layers.
          */
-        this.setPointerMetaLimits = core.setPointerMetaLimits;
+        this.setPointerMetaLimits = function (node, name, min, max) {
+            ensureNode(node, 'node');
+            ensureType(name, 'name', 'string');
+            ensureMinMax(min, 'min');
+            ensureMinMax(max, 'max');
+
+            return core.setPointerMetaLimits(node, name, min, max);
```
pmeijer: This should be checked for starting with '_', 'ovr', 'base' and 'member'.

src/common/core/core.js
```diff
@@ -1695,25 +2808,45 @@ define([
...
+         * @throws {CoreIllegalOperationError} If the context of the operation is not allowed.
+         * @throws {CoreAssertError} If some internal error took place inside the core layers.
          */
-        this.setPointerMetaTarget = core.setPointerMetaTarget;
+        this.setPointerMetaTarget = function (node, name, target, min, max) {
+            ensureNode(node, 'node');
+            ensureType(name, 'name', 'string');
+            ensureNode(target, 'target');
+            ensureMinMax(min, 'min');
+            ensureMinMax(max, 'max');
```
pmeijer: This should be checked for starting with '_', 'ovr', 'base' and 'member'.

src/common/core/CoreAssertError.js
```diff
@@ -0,0 +1,23 @@
+/*globals define*/
+/*jshint node: true, browser: true*/
+
+/**
+ * @author kecso / https://github.com/kecso
+ */
+
+define([], function () {
```
pmeijer: CoreInternalError

PR #1263, opened 2016-12-18
Detect SIGINT from posix exit code. Fixes #1262

### LINKED ISSUE #1262: Worker doesn't detect SIGINT as exit code 130 on *nix
When the job for the worker requires cleanup before exiting, the job for the worker can override the exit event handlers and the `SIGINT` event (in case it is canceled from webgme):

```
process.on('SIGINT', function() {
    // Clean up code for the job
});
```

However, this can lead to a problem since calling `process.exit` will no longer show the job was canceled using `SIGINT` (the worker checks the signal to see if the job was canceled which will be null).

In *nix, exit code 130 corresponds to `SIGINT`; recognizing this in the worker should solve this problem.

Some examples of failed work-arounds:
```
process.on('SIGINT', function() {
    process.exit(130);  // returns exitCode=130; signal=null
});

process.on('SIGINT', function() {
    process.exit('SIGINT');  // returns exitCode=0; signal=null
});

process.on('SIGINT', function() {
    process.exit();  // returns exitCode=0; signal=null
});
```

brollb:


pmeijer:
Fixed in [PR-2 in executor-worker](https://github.com/webgme/executor-worker/pull/2)

PR #1264, opened 2016-12-19
Use a different template for the generated source code documentation

kecso:
This enhancement changes visuality of the source documents.
Now it is easier to find functions.
Also the documents have been trimmed to fit better to those API that are actually public or intended to use by users of WebGME.
Small fixes are also covered for client and core.


pmeijer:
I think the added client tests aren't being run by travis nor appveyor.

PR #1283, opened 2017-01-07
Support copy to clipboard at model export

### LINKED ISSUE #1272: SaveToDisk failing w/ blob url
Using `SaveToDisk` to download using the blob url results in nothing happening... It seems to be caused by  https://github.com/webgme/webgme/blob/master/src/client/js/Utils/SaveToDisk.js#L23-L24.

Why not simply call `save.click()`?

kecso:
Whenever the user exports some model, the notification tab will provide an option to copy its blobhash to the clipboard. This hash can be later used when the user wants to import the model into another project. The blob has to exist in the target deployment's blobStorage, so it is best if used within the same deployment.
Also this enhancement implements the BlobHashWidget which is an alternate version of the AssetWidget as it accepts a hash string and if matches it provides the link to the asset (just as the AssetWidget, after the upload is successful).

Fixes #1272 SaveToDisk failing w/ blob url


src/client/js/Controls/PropertyGrid/Widgets/BlobHashWidget.js
```diff
@@ -0,0 +1,174 @@
...
+            bytes = bytes / thresh;
+            u += 1;
+        } while (bytes >= thresh);
+
+        return bytes.toFixed(1) + ' ' + units[u];
+    };
+
+    BlobHashWidget.prototype.destroy = function () {
+        this._detachHandlers();
+        clearTimeout(this.__timeoutId);
```
pmeijer: Where is the __timeoutId coming from?

src/client/js/main.js
```diff
@@ -68,6 +68,7 @@ require.config({
         d3: 'bower_components/d3/d3.min',
         epiceditor: 'bower_components/EpicEditor/epiceditor/js/epiceditor',
         ravenjs: 'bower_components/raven-js/dist/raven',
+        clipboard: 'bower_components/clipboard/dist/clipboard.min',
```
pmeijer: There's no non-minified version?
The lib distribution should also be updated.

src/client/js/Utils/Exporters.js
```diff
@@ -4,7 +4,11 @@
 /**
  * @author kecso / https://github.com/kecso
  */
-define(['js/Loader/ProgressNotification'], function (ProgressNotification) {
+define([
+    'js/Loader/ProgressNotification',
+    'clipboard',
+    'js/Utils/SaveToDisk'
+], function (ProgressNotification, Clipboard, saveToDisk) {
```
pmeijer: Can you remove the unused Notification Manager and the gritter bower component?

src/client/js/Controls/PropertyGrid/Widgets/BlobHashWidget.js
```diff
@@ -0,0 +1,174 @@
...
+ * @author kecso / https://github.com/kecso
+ */
+
+define([
+    'js/Controls/PropertyGrid/Widgets/WidgetBase',
+    'blob/BlobClient',
+    'js/logger',
+    'common/regexp',
+    'css!./styles/BlobHashWidget.css'
+], function (WidgetBase,
```
pmeijer: It would have been a bit more elegant if this could share some implementation with the AssetWidget instead of copying the common code.

src/client/js/Utils/Exporters.js
```diff
@@ -51,6 +56,7 @@ define(['js/Loader/ProgressNotification'], function (ProgressNotification) {
             selectedIds,
             withAssets,
             function (err, result) {
+                var copied = false;
```
pmeijer: Used?

src/client/js/mainDEBUG.js
```diff
@@ -49,6 +48,7 @@ require.config({
         d3: 'bower_components/d3/d3',
         epiceditor: 'bower_components/EpicEditor/epiceditor/js/epiceditor',
         ravenjs: 'bower_components/raven-js/dist/raven',
+        clipboard: 'bower_components/clipboard/dist/clipboard.min',
```
pmeijer: Better to use non-minified in debug mode

src/client/js/Utils/Exporters.js
```diff
@@ -60,12 +65,21 @@ define(['js/Loader/ProgressNotification'], function (ProgressNotification) {
...
+                        type: 'success',
+                        icon: 'glyphicon glyphicon-copy'
                     });
+
+                    //TODO it is not the nicest way, we might want to add it as an option to the notification itself
+                    $(progress.note.$ele).find('.glyphicon-copy').attr('title', 'copy blobhash of export to clipboard');
+                    $(progress.note.$ele).find('.glyphicon-copy').attr('data-clipboard-text', result.hash);
+                    $(progress.note.$ele).find('.glyphicon-copy').css('cursor', 'copy');
+                    $(progress.note.$ele).find('.glyphicon-copy').addClass('btn btn-xs');
+                    new Clipboard($(progress.note.$ele).find('.glyphicon-copy')[0]);
```
pmeijer: Minor note, instead of finding the icon five times you could either make a local variable or chain the setters.

pmeijer: Would it be possible to add the Clipboard handling to the entire notification?

src/client/js/Utils/Exporters.js
```diff
@@ -60,12 +65,21 @@ define(['js/Loader/ProgressNotification'], function (ProgressNotification) {
                         progress: 100
                     });
                 } else {
+
                     progress.note.update({
                         message: '<strong>Exported </strong> models <a href="' +
                         result.downloadUrl + '" target="_blank">' + result.fileName + '</a>',
                         progress: 100,
-                        type: 'success'
+                        type: 'success',
```
pmeijer: `(click to copy hash to clipboard)`

PR #1286, opened 2017-01-11
Drag-n-Drop onto designer items adds drag-items to set.

kecso:
With this enhancement, the user will be able to drag and drop members into the owner of the set (just like setting the target of pointers worked).

It is functional in the model- and svg-decorator. The drop accept mechanisms are implemented in the base class, so other decorators can use it as well.

Also a new icon in the model decorator highlights if the node has defined sets. Double clicking it brings up the set editor view.


src/client/decorators/ModelDecorator/Core/_ModelDecorator.scss
```diff
@@ -126,6 +126,33 @@ $abstract-class-name-color: #AAAAAA;
       }
     }
 
+    .set {
+      position: absolute;
+      bottom: 0;
```
pmeijer: This could've been combined with ptr, but no biggie..

PR #1292, opened 2017-01-11
Upgrading seeds so they use the latest data format.

kecso:


test/client/js/client/gmeNodeGetter.spec.js
```diff
@@ -314,13 +314,13 @@ describe('gmeNodeGetter', function () {
     it('should return the value of the given registry', function () {
         var node = getNode('/175547009/871430202', logger, basicState, basicStoreNode);
 
-        expect(node.getRegistry('position')).to.eql({x: 91, y: 219});
+        expect(node.getRegistry('position')).to.eql({x: 104, y: 218});
```
pmeijer: The SigalFlowSystem still has Meta-inconsistencies. Also the SignalFlowModel has meta defined - either add it to the meta or remove it.
Check the meta-rules for the seeds too before porting.

PR #1298, opened 2017-01-16
Add option to point to a custom plugin configurator

pmeijer:
The `metadata.json` associated with a plugin now takes an optional parameter, `configWidget`, with a path to an AMD (requirejs module). If defined the InterpeterManager in the generic UI will call that instead of the default config dialog. 

For API of configurator see `src/plugin/coreplugins/CustomPluginConfig/widget/ConfigWidget.js`.

Note that you can reuse the default dialog by including and calling it from your custom widget.


src/client/js/Utils/InterpreterManager.js
```diff
@@ -45,77 +45,105 @@ define([
...
+    InterpreterManager.prototype.loadConfigDialog = function (metadata, callback) {
+        if (!metadata.configWidget) {
+            callback(null, PluginConfigDialog);
+        } else {
+            requirejs([metadata.configWidget],
+                function (CustomConfigDialog) {
+                    callback(null, CustomConfigDialog);
+                },
+                function (err) {
+                    callback(err);
```
kecso: Couldn't it fall back to the PluginConfigDialog?
Maybe with an extra notification...

src/plugin/coreplugins/CustomPluginConfig/widget/ConfigWidget.js
```diff
@@ -0,0 +1,52 @@
+/*globals, define, $, WebGMEGlobal*/
```
kecso: A base class would be nice.

src/client/js/Utils/InterpreterManager.js
pmeijer: I think it's better fail hard here. The plugin might be unusable without the custom configuration.

src/plugin/coreplugins/CustomPluginConfig/widget/ConfigWidget.js
pmeijer: The API is so small and the base class wouldn't really provide any functionality.

PR #1299, opened 2017-01-16
Library notification

kecso:
If a library is used from a URL (meaning that a branch is followed and the project is in the same deployment), we can notify the user if a new version is available.

### DIFF
--- a/src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
+++ b/src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
@@ -46,10 +46,11 @@ define(['js/logger',
             stateLoaded = 1,
             selfId,
             selfPatterns = {},
-        //local container for accounting the currently opened node list,
-        // its a hashmap with a key of nodeId and a value of { FancyTreeNode, childrenIds[], state }
+            //local container for accounting the currently opened node list,
+            // its a hashmap with a key of nodeId and a value of { FancyTreeNode, childrenIds[], state }
             nodes = {},
             refresh,
+            checkLibraries,
             initialize,
             initialized = false,
             self = this,
@@ -159,7 +160,7 @@ define(['js/logger',
                 c = isMetaNode ? GME_META_MODEL_CLASS : GME_MODEL_CLASS;
             } else {
                 //by default everyone is represented with the atom class
-                c = isMetaNode ? GME_META_ATOM_CLASS :GME_ATOM_CLASS;
+                c = isMetaNode ? GME_META_ATOM_CLASS : GME_ATOM_CLASS;
             }
 
             return c;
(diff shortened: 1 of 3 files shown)


src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
```diff
@@ -592,6 +598,11 @@ define(['js/logger',
                     updatedObject = client.getNode(objectId);
 
                     if (updatedObject) {
+
+                        if (objectId === CONSTANTS.PROJECT_ROOT_ID) {
+                            checkLibraries();
```
pmeijer: Won't this notify at every change to the root-node? That is every change..

src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
```diff
@@ -783,6 +794,29 @@ define(['js/logger',
...
+                    self._libraryChecks[name] = true;
+                    self._libraryManager.check(name,
+                        function (err, notified) {
+                            if (!err && !notified) {
+                                delete self._libraryChecks[name];
+                            }
+                        });
+                };
+
+            self._libraryChecks = self._libraryChecks || {};
```
pmeijer: Oh I see that the check is being book-kept here.. However is this state ever reset?

pmeijer:
Pushing this to next release..

src/client/js/Panels/ObjectBrowser/TreeBrowserControl.js
pmeijer: Can't you just check it at the time where a library node is created in the tree?

src/client/js/Utils/LibraryManager.js
```diff
@@ -108,5 +126,170 @@ define([
...
+                        callback(null, true);
+                    } else {
+                        callback(null, false);
+                    }
+                });
+            }
+        }
+    };
+
+    LibraryManager.prototype._clearWatches = function () {
```
pmeijer: I think you mean watchers.

src/client/js/Utils/LibraryManager.js
```diff
@@ -108,5 +126,170 @@ define([
...
+            delete this._followedProjects[projectId];
+        }
+    };
+
+    LibraryManager.prototype._watchProject = function (projectId) {
+        var self = this,
+            eventFunction = function (websocket, event) {
+                var libraryName;
+
+                if (event && event.etype && event.etype === CONSTANTS.BRANCH_HASH_UPDATED &&
```
pmeijer: Two first checks are redundant - an event is always returned and event.projectId too.

src/client/js/Utils/LibraryManager.js
```diff
@@ -108,5 +126,170 @@ define([
...
+                            commitHash: info.commitHash,
+                            notified: false
+                        };
+                        this._checkLibrary(availableNames[i]);
+                    }
+                }
+            }
+
+        } else {
+
```
pmeijer: // Project and/or branch changed -> reset the library watchers.

PR #1305, opened 2017-01-23
Version and add bundled dist files to browsers' app-cache.

pmeijer:
This PR introduces app-caching as described here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache

The files being cached are the bundled distribution files located at (`/dist`). These are pretty large files and loaded step-by-step to ensure that the dependencies are loaded first. This however slows down the loading process as those request cannot be made in parallel - thus caching can significantly speed things up.

image

To ensure that wrong versions aren't accidentally used (at the point of cache re-evaluation) generated dist-files are now version controlled using the npm version.

N.B. In the link above it says that app-cache is deprecated and will be replaced by:
https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers

However the later is still experimental has poorer browser support. At a transition point, all that needs to be removed for the app-cache is the manifest tag in the index.html. Also note that this PR does not try to solve the working off-line problem - but rather speed up initial loading of large bundles for slow connections.

### DIFF
--- a/src/client/index.html
+++ b/src/client/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html manifest="webgme.dist.appcache">
 <head>
     <title>WebGME</title>
     <!-- Load main.js with RequireJS and launch the app from there -->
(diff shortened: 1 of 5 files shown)


pmeijer:
N.B. After merging this - make sure to bump the webgme version in the master branch to `2.10.0-beta1`

PR #1309, opened 2017-01-30
Indicate that a type is a connection on meta-editor

pmeijer:
image


kecso:
Shouldn't we put this visual feature under some sort of configuration?
I know that we already refer to src+dst as an implicit definition of a connection (maybe we should put those names behind a configuration as well), but right now the Meta editor does not have any internal interpretation and it seems another step towards hardwiring src/dst. (on the other hand the default behavior should be to use this visual feature)

PR #1317, opened 2017-02-07
Load members shorthand function

kecso:
The enhancement implements the loadMembers and loadOwnMembers shorthand functions on the core API. They collect the members of a given set of the given node.
Also it takes care of the missing handling of set registry and set attribute changes (during diff generation, conflict management and merging).

### DIFF
--- a/src/common/core/core.js
+++ b/src/common/core/core.js
@@ -1727,7 +1727,7 @@ define([
             ensureNode(node, 'node');
             ensureType(name, 'name', 'string');
             var names = core.getSetNames(node).concat(core.getValidSetNames(node));
-            ;
+
             if (names.indexOf(name) === -1) {
                 throw new CoreIllegalOperationError('Cannot access member information of unknown set.');
             }
--- a/src/common/core/core.js
+++ b/src/common/core/core.js
@@ -1726,7 +1726,8 @@ define([
         this.getOwnMemberPaths = function (node, name) {
             ensureNode(node, 'node');
             ensureType(name, 'name', 'string');
-            var names = core.getSetNames(node).concat(core.getValidSetNames(node));;
+            var names = core.getSetNames(node).concat(core.getValidSetNames(node));
+            ;
             if (names.indexOf(name) === -1) {
                 throw new CoreIllegalOperationError('Cannot access member information of unknown set.');
             }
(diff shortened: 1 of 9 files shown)


src/common/core/core.js
```diff
@@ -1726,7 +1726,8 @@ define([
         this.getOwnMemberPaths = function (node, name) {
             ensureNode(node, 'node');
             ensureType(name, 'name', 'string');
-            var names = core.getSetNames(node).concat(core.getValidSetNames(node));;
+            var names = core.getSetNames(node).concat(core.getValidSetNames(node));
+            ;
```
pmeijer: Colon

src/common/core/corediff.js
```diff
@@ -1183,12 +1192,37 @@ define(['common/util/canon',
                     self.deleteSet(node, setNames[i]);
                 } else {
                     self.createSet(node, setNames[i]);
+                    if ((Object.keys(setDiff[setNames[i]].attr || {})).length > 0) {
```
pmeijer: Why the extra parenthesis?

src/common/core/corediff.js
```diff
@@ -185,7 +185,16 @@ define(['common/util/canon',
...
-                        data[names[i]] = {};
+                        data[names[i]] = {attr: {}, reg: {}};
+                        keys = self.getOwnSetAttributeNames(node, names[i]);
+                        for (j = 0; j < keys.length; j += 1) {
+                            data[names[i]].attr[keys[j]] = self.getSetAttribute(node, names[i], keys[j]);
+                        }
+                        keys = self.getOwnSetRegistryNames(node, names[i]);
+                        for (j = 0; j < keys.length; j += 1) {
+                            data[names[i]].reg[keys[j]] = self.getSetRegistry(node, names[i], keys[j]);
+                        }
```
pmeijer: These could be getOwnSetAtr/Reg. Not sure if it really matter though..

src/common/core/setcore.js
```diff
@@ -278,6 +282,20 @@ define(['common/core/CoreAssert', 'common/core/constants'], function (ASSERT, CO
...
+        function loadNodesOfPaths(root, paths) {
+            var nodes = [],
+                i,
+                rootHash = self.getHash(root);
+
+            return TASYNC.call(function () {
+                for (i = 0; i < paths.length; i += 1) {
+                    nodes[i] = self.loadByPath(root, paths[i]);
+                }
+                return TASYNC.lift(nodes);
```
pmeijer: Couldn't this suffer from the same issue as with children?
That they have been removed and loadByPath returns null?
I think it doesn't since the members are stored in the ovr and cleaned at deletion.. But just checking..

PR #1329, opened 2017-02-22
added typings into the project

phreed:
This is the same set of typings previously available separately.
When this has been verified those other locations will be deprecated.


pmeijer:
This is a duplicate of #1341

Issue #1089, opened 2016-08-02
Option to hide/filter nodes in Part-Browser when root is opened.


pmeijer:
Request from Nag

pmeijer:
This can be achieved by registering a specific visualizer at the root-node that implements `getValidTypesInfo` and returns an empty object. See #1332

PR #1332, opened 2017-02-28
Allow visualizer to defined getValidTypesInfo to list part-browser nodes.

### LINKED ISSUE #1331: Deletion of the last Cross Cut tab leaves the members on the canvas.

pmeijer:
If `Panel.getValidTypesInfo` is not defined and a function - the containment behavior based on the meta-rules is applied.

This gets rid of the hard-coding of visualizers in the part-browser and enables visualizers to defined their own behavior based on the context.

This PR also fixes #1331


pmeijer:
Related to #1089

PR #1333, opened 2017-03-07
Shard large overlays of core data-blobs

kecso:
The feature introduces sharding of bigger overlay objects.
If the outmost level entries of the overlays of any node reaches the half-point of shardSize, the core transforms the single overlay object into a sharded overlay.
The sharded overlay is a collection of relid driven, hashed objects that contains the actual overlay entries.
A single source will only occur in a single shard, no matter how many outgoing relations it has.
Items in a shard are counted on relation level, not on source level. Once the item amount reaches the limit, no new source entries will be added allowed, but the already existing ones can still get new relations.
Once an overlay becomes sharded, it will never go back to single overlay.
The change aims to overcome issues when the root object becomes too big to store as a single object in mongodb, and the hash computation after each change became visibly slow.

With a sharded overlay, the following changes in time were measured (50k object project size):
single registry update: 0.7s -> 0.04s
copy 50k nodes: 14.5s -> 2.2s
delete 50k nodes: 31s -> 16.13s


src/common/core/corerel.js
```diff
@@ -125,6 +133,258 @@ define([
...
+            return (self.getProperty(node, CONSTANTS.OVERLAYS_PROPERTY) || {})[CONSTANTS.OVERLAY_SHARD_INDICATOR]
+                === true;
+        }
+
+        function shouldHaveShardedOverlays(node) {
+            return Object.keys(self.getProperty(node, CONSTANTS.OVERLAYS_PROPERTY) || {}).length >=
+                _shardingLimit && self.getPath(node).indexOf('_') === -1;
+        }
+
+        function reserveOverlayShard(node) {
```
pmeijer: Why reserve instead of createNew/addNew?

src/common/storage/project/cache.js
```diff
@@ -89,17 +89,36 @@ define([
...
-                                    ASSERT(missing[key] === obj);
-
-                                    delete missing[key];
-                                    if (!err && obj2) {
-                                        cacheInsert(key, obj2);
+                                var callbacks,
+                                    subKey,
+                                    cb;
+
+                                if ((obj2 || {}).hasOwnProperty('multipleObjects') && obj2.multipleObjects === true) {
```
pmeijer: `(obj2 || {}).multipleObjects === true` is the same.

pmeijer: Maybe rename `obj2` at this point to maybe `loaded` at this point.

pmeijer: Actually this whole method could use some better naming.. 
Like a separate variable for the missing-callbacks.

config/validator.js
```diff
@@ -148,6 +148,10 @@ function validateConfig(configOrFileName) {
     // core
     expectedKeys.push('core');
     assertBoolean('config.core.enableCustomConstraints', config.core.enableCustomConstraints);
+    assertNumber('config.core.overlayShardSize', config.core.overlayShardSize);
+    if(config.core.overlayShardSize < 1000){
+        throw new Error('Overlay shard size must be at least 1000.');
+    }
```
kecso: Mainly it is there not to set a negative or zero number, but then I thought probably setting it too low would do more harm at the end for the user.
The tests are not validating whenever they change these values.

src/server/storage/storagehelpers.js
```diff
@@ -89,6 +141,17 @@ function loadPath(dbProject, rootHash, loadedObjects, path, excludeParents) {
     var deferred = Q.defer(),
         pathArray = path.split('/');
 
+    function processLoadResult(hash, result) {
+        var subHash;
+        if (result.hasOwnProperty('multipleObjects') && result.multipleObjects === true) {
```
pmeijer: `result.multipleObjects === true` suffices.

src/server/storage/storagehelpers.js
```diff
@@ -89,6 +141,17 @@ function loadPath(dbProject, rootHash, loadedObjects, path, excludeParents) {
...
+            for (subHash in result.objects) {
+                loadedObjects[subHash] = result.objects(subHash);
```
pmeijer: Check the coverage of these:
image

src/common/core/corerel.js
```diff
@@ -125,6 +133,258 @@ define([
...
+            self.setProperty(self.getChild(node, CONSTANTS.OVERLAYS_PROPERTY), CONSTANTS.OVERLAY_SHARD_INDICATOR, true);
+
+            node.overlays = {};
+            node.overlayMutations = {};
+            node.overlayInitials = {};
+            node.minimalOverlayShardId = null;
+            for (source in originalOverlays) {
+                if (source !== CONSTANTS.MUTABLE_PROPERTY) {
+                    for (name in originalOverlays[source]) {
+                        if (name !== CONSTANTS.MUTABLE_PROPERTY) {
```
kecso: yes it does... we handle the 'source' as a child, not just the overlays...

src/common/core/corerel.js
```diff
@@ -125,6 +133,258 @@ define([
...
+
+            for (shardId in node.overlays) {
+                if (node.overlays[shardId].items[source] &&
+                    typeof node.overlays[shardId].items[source][name] === 'string') {
+
+                    ensureOverlayShardMutated(node, shardId);
+
+                    delete node.overlays[shardId].items[source][name];
+                    node.overlays[shardId].itemCount -= 1;
+                    break;
```
kecso: The source entry is being removed during overlayPersist.
If the whole shard become empty, it will not be removed right away, but if it remains empty for the next commit as well..

src/common/core/corerel.js
```diff
@@ -174,15 +458,26 @@ define([
...
-                    target = overlay[source][name];
-                    inverseOverlays[target] = inverseOverlays[target] || {};
-                    inverseOverlays[target][name] = inverseOverlays[target][name] || [];
-                    inverseOverlays[target][name].push(source);
+            for (shardId in overlaysObject) {
+                overlay = overlaysObject[shardId];
+                for (source in overlay.items) {
+                    if (source !== CONSTANTS.MUTABLE_PROPERTY) {
+                        for (name in overlay.items[source]) {
+                            if (name !== CONSTANTS.MUTABLE_PROPERTY) {
```
kecso: unfortunately it does here as well (or more precisely it could).

src/common/util/jsonPatcher.js
```diff
@@ -269,12 +357,25 @@ define([
...
-            for (key in targetJson.ovr) {
-                if (sourceJson.ovr[key]) {
-                    patch = patch.concat(
-                        diff(sourceJson.ovr[key], targetJson.ovr[key], '/ovr/' + _strEncode(key) + '/', [], false, key)
-                    );
+            if (sourceJson.ovr.sharded !== true && targetJson.ovr.sharded === true) {
+                // Transformation into sharded overlays means that we have to collect update information.
+                patch.push(getShardingDiff(sourceJson.ovr, targetJson.ovr));
+            } else if (sourceJson.ovr.sharded === true && targetJson.ovr.sharded === true) {
+                patch = patch.concat(diff(sourceJson.ovr, targetJson.ovr, '/ovr/', [], false, '', true, false));
```
pmeijer: Should noUpdate be false here? It's not in the regular case..

src/common/util/jsonPatcher.js
```diff
@@ -202,19 +199,104 @@ define([
...
+        if (isAddition) {
+            patchItem.value = items;
+        } else {
+            patchItem.value = {};
+        }
+
+        for (source in items) {
+            patchItem.updates.push(source);
+            for (name in items[source]) {
+                patchItem.partialUpdates.push(items[source][name]);
```
pmeijer: Couldn't this include /_nullptrs?

pmeijer: In the `diff` function this means an update on the owner of the nulltptr.

src/common/util/jsonPatcher.js
```diff
@@ -430,6 +531,179 @@ define([
...
+            return false;
+        }
+
+        if (patchItem.op === 'add' || patchItem.op === 'replace') {
+            return REGEXP.HASH.test(patchItem.value);
+        }
+    }
+
+    function _getChangedNodesFromShard(patch, res, hash, gmePath) {
+        var shardPatch = patch[hash] && patch[hash].patch ? patch[hash] && patch[hash].patch : patch[hash],
```
pmeijer: `patch[hash] && patch[hash].patch ? patch[hash].patch : patch[hash]`

src/common/util/jsonPatcher.js
kecso: no.
That case is when the '/ovr' of the coreNode contains only hashes with relids. In the regular case, we deal with the updates of a source entry later (to get the detailed relation changes). In the sharded case, we do not have any more detail (at least not in the regular coreNode), so the updates should be generated as well at that point.

src/common/core/corerel.js
```diff
@@ -655,14 +658,14 @@ define([
                 }
             }
 
-            //Now we check if we need to mutate the inverse overlays
+            //First check if we the inverse computed already and if we need to mutate it
```
pmeijer: ?

PR #1338, opened 2017-03-10
Remove global state in core diff for async method and add tests cases.

pmeijer:


src/common/core/corediff.js
```diff
@@ -2511,6 +2518,9 @@ define(['common/util/canon',
         //</editor-fold>
 
         //<editor-fold=Added Methods>
+
+        // FIXME: It really looks like the diff requires that no nodes are mutated. This must be documented somewhere.
+        // FIXME: Maybe checking for isMutated in core.js at the roots are enough..
```
kecso: yeah, that check should suffice
on the other hand, it is not completely true that you need to have 'not mutated' trees to compare. All the data is interpreted, so it should be possible with any kind of trees - though we still use the hashes of children to figure out if we should go deeper, but that could be changed...

src/common/core/corediff.js
```diff
@@ -2557,12 +2566,19 @@ define(['common/util/canon',
...
+         *
+         * @returns {object}
+         */
         this.tryToConcatChanges = function (base, extension) {
             var result = {};
             _conflictItems = [];
             _conflictMine = {};
             _conflictTheirs = {};
-            _concatBase = base;
+            _concatBase = base; // FIXME: Are these meant to be copies, currently base and extension are mutated?
```
kecso: I would say, that it is not an issue because of the current use-cases (the base input is never used after this, rather the result is the one that will be ultimately used), but it would be probably wiser to have the _concatBase as a deep copy and not mutate any input objects.

src/common/core/corediff.js
```diff
@@ -111,31 +118,33 @@ define(['common/util/canon',
...
-                i,
-                diff = {};
+                diff = {},
+                sReg,
+                tReg,
+                i;
 
             for (i = 0; i < sNames.length; i++) {
                 if (tNames.indexOf(sNames[i]) === -1) {
+                    // FIXME: This constant should be renamed to something with . or $ (which is illegal in mongo)
```
pmeijer: Actually it struck me that $ is valid as a value of attr/reg etc.

src/common/core/corediff.js
```diff
@@ -94,14 +102,13 @@ define(['common/util/canon',
...
-                    diff[tNames[i]] = self.getAttribute(target, tNames[i]);
-                } else {
-                    if (CANON.stringify(self.getAttribute(source, tNames[i])) !==
-                        CANON.stringify(self.getAttribute(target, tNames[i]))) {
+                sAttr = self.getOwnAttribute(source, tNames[i]);
+                tAttr = self.getOwnAttribute(target, tNames[i]);
 
-                        diff[tNames[i]] = self.getAttribute(target, tNames[i]);
-                    }
+                if (sAttr === undefined) {
```
pmeijer: True

src/common/core/corediff.js
```diff
@@ -235,6 +244,7 @@ define(['common/util/canon',
 
         function metaDiff(source, target) {
             //TODO jsonMeta format should be changed in all places!!!
+            // FIXME: which places??
```
pmeijer: Oh I see, I removed the todo

PR #1341, opened 2017-03-13
Add Typescript bindings

phreed:
This includes the most recent version of the API type definitions.
These are primarily useful in two situations:
1) When writing a TypeScript plugin (or some extension to webGME)
2) When using an IDE that has intellisense (VisualStudioCode, WebStorm, Atom, etc.)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
pmeijer: Can you add .vscode to .gitignore?

pmeijer: nm, I'll fix it in the prerelease branch

PR #1343, opened 2017-03-19
Sharded copy fix

kecso:
When copying a node with overlay shards, the copy should also contain the (in memory) shards-info: `overlays`, `overlayMutations` and `minimalOverlayShardId`.

### DIFF
--- a/src/common/core/corerel.js
+++ b/src/common/core/corerel.js
@@ -944,6 +944,7 @@ define([
             newNode.inverseOverlaysMutable = node.inverseOverlaysMutable;
 
             if (hasShardedOverlays(node)) {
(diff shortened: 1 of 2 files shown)


src/common/core/corerel.js
```diff
@@ -943,6 +943,12 @@ define([
             }
             newNode.inverseOverlaysMutable = node.inverseOverlaysMutable;
 
+            if (hasShardedOverlays(node)) {
+                newNode.overlays = JSON.parse(JSON.stringify(node.overlays));
```
pmeijer: `// Copy the shards-info for nodes with sharded overlay #1343`

PR #1344, opened 2017-03-20
Fixes potential GUID collisions among copied nodes

kecso:
As a result of having shorter relids, the possible collision of node GUIDs (that only differ in their places in the containment tree, but copies of each other) increased.
Because in a majority of copy use-cases, the node is modified as well (for example getting a new position) the structural reuse would not work anyhow, so we generate a new GUID for the copy to avoid collisions.

As it is still unlikely that existing projects have colliding GUIDs (it would require a great amount of copies on from and to specific levels of the containment), we do not require checks for existing projects, but we provide a plugin, `GuidCollider`, that can find and, if needed, eliminate collisions.

### DIFF
--- a/src/common/core/guidcore.js
+++ b/src/common/core/guidcore.js
@@ -192,6 +192,7 @@ define([
         this.copyNode = function (node, parent) {
             var newNode = innerCore.copyNode(node, parent);
 
+            // Generate new guid at copy #1344
             setDataGuid(newNode, GUID());
 
             return newNode;
@@ -201,6 +202,7 @@ define([
             var newNodes = innerCore.copyNodes(nodes, parent),
                 i;
 
+            // Generate new guids at copy #1344
             for (i = 0; i < newNodes.length; i += 1) {
                 setDataGuid(newNodes[i], GUID());
             }
--- a/src/common/core/guidcore.js
+++ b/src/common/core/guidcore.js
@@ -188,6 +188,25 @@ define([
                 return CONSTANTS.NULL_GUID;
             }
         };
+
+        this.copyNode = function (node, parent) {
+            var newNode = innerCore.copyNode(node, parent);
+
+            setDataGuid(newNode, GUID());
+
+            return newNode;
+        };
+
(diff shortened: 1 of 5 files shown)


src/plugin/coreplugins/GuidCollider/GuidCollider.js
```diff
@@ -0,0 +1,101 @@
+/*globals define*/
+/*jshint node:true, browser:true*/
+
+/**
+ * Plugin mainly used for testing.
+ * @author lattmann / https://github.com/lattmann
```
pmeijer: author..

src/common/core/guidcore.js
```diff
@@ -188,6 +188,25 @@ define([
                 return CONSTANTS.NULL_GUID;
             }
         };
+
+        this.copyNode = function (node, parent) {
+            var newNode = innerCore.copyNode(node, parent);
+
```
pmeijer: `// Generate new guid at copy #1344`

src/common/core/guidcore.js
```diff
@@ -188,6 +188,25 @@ define([
...
+            setDataGuid(newNode, GUID());
+
+            return newNode;
+        };
+
+        this.copyNodes = function (nodes, parent) {
+            var newNodes = innerCore.copyNodes(nodes, parent),
+                i;
+
```
pmeijer: `// Generate new guids at copy #1344`

src/plugin/coreplugins/GuidCollider/GuidCollider.js
```diff
@@ -0,0 +1,101 @@
...
+                }
+                next(null);
+            };
+
+        self.updateMETA(self.metaTypes);
+
+        Q.ninvoke(self.core, 'traverse', self.rootNode, null, visitFn)
+            .then(function () {
+                if (hasCollision && currentConfiguration.checkOnly === false) {
+                    return Q.ninvoke(self, 'save', 'guid collision fix');
```
pmeijer: FYI: save returns a promise already

PR #1348, opened 2017-03-22
Closes #1169 Extra svgs are linked to and served from their actual location.

### LINKED ISSUE #1169: Sometimes when starting server it never comes up
When it fails, there are no errors and it does not crash, you just never see the `Server is listening...` message so you cannot connect. After a random amount of `ctrl+c` & restarting the server (i.e. restarting the server a random amount of times) at some point the message will show and you'll be able to connect to the server.  No errors are produced.

pmeijer:
This gets rid of the bad copying and generating svg list at server start up.

Additionally the svg picker only shows the default in the first tab (and not all available svgs).


src/utils.js
```diff
@@ -9,10 +9,40 @@
...
 var fs = require('fs'),
     Q = require('q'),
-    genDecoratorSvgList = require('./client/assets/generate_decorator_svg_list'),
-    ncp = require('ncp'), // Module for copying entire directory
     path = require('path'),
-    requireUncached = require('require-uncached');
+    requireUncached = require('require-uncached'),
+    SVGMapDeffered;
+
+function walkDir (dir, done) {
```
kecso: why not using promise form?

src/utils.js
```diff
@@ -52,39 +82,56 @@ function getComponentNames(basePaths) {
...
-/**
- * @param gmeConfig
- * @param logger
- * @param callback
- * @returns {*}
- */
-function copySvgDirsAndRegenerateSVGList(gmeConfig, logger, callback) {
-    var deferred = Q.defer(),
-        svgAssetDir = path.join(__dirname, 'client', 'assets', 'DecoratorSVG');
+function getSVGMap(gmeConfig, logger, callback) {
```
kecso: couldn't it be even synchronous?
I know it would be blocking, but we do not expect too deep and too large sets of files (so it should be fairly fast).

src/utils.js
pmeijer: I borrowed it from the `generate_decorator_svg.list` and didn't convert it. Will do though

src/utils.js
pmeijer: The amount of extra svgs is user specific and not bounded. Don't see why it should block?

src/utils.js
```diff
@@ -52,39 +91,56 @@ function getComponentNames(basePaths) {
...
-        .then(function () {
-            return genDecoratorSvgList();
+    if (SVGMapDeffered) {
+        return SVGMapDeffered.promise.nodeify(callback);
+    }
+
+    SVGMapDeffered = Q.defer();
+    svgMap = {};
+
+    Q.nfcall(walkDir, svgAssetDir)
```
kecso: promisified already

src/utils.js
kecso: it was a weak try :)
so async it is

src/utils.js
```diff
@@ -3,16 +3,55 @@
...
+        if (!pending) {
+            deferred.resolve(results);
+            return;
+        }
+
+        list.forEach(function (file) {
+            file = path.join(dir, file);
+            fs.stat(file, function (err, stat) {
+                if (stat && stat.isDirectory()) {
+                    walkDir(file, function (err, res) {
```
kecso: promisified (and for clarity, probably the fs.stat could be called as a promise)

finger563:
upgrading from 2.11 to 2.12 seems to have messed up the paths in my models to the svgs, is this expected?

pmeijer:
Will it's not expected.. I'm looking into it right now. Thanks for the heads up!

pmeijer:
I fixed it in a branch, was surprised at first that the test didn't catch it... 
So the issue is having relative paths on ubuntu. A workaround till the fix is out is to do:
`path.join(__dirname, '..', <svgDir>)`

finger563:
Thanks that fixed it :)

PR #1359, opened 2017-04-05
Copy nodes

kecso:
The copyNodes function have been reimplemented and corrected.
Now, it is able to follow the chain if inheritance correctly when preserving 'copy-internal' relationships.
Also it is provided on the client in its 'purest' form to hopefully replace the copyMoreNodes at some point in the futue.

### DIFF
--- a/src/common/core/corerel.js
+++ b/src/common/core/corerel.js
@@ -956,49 +956,126 @@ define([
         };
 
         this.copyNodes = function (nodes, parent, takenRelids, relidLength) {
-            //copying multiple nodes at once for keeping their internal relations
-            var paths = [],
-                i, j, index, names, pointer, newNode,
-                copiedNodes = [],
-                // Every single element will be an object with the
-                // internally pointing relations and the index of the target.
-                internalRelationPaths = [];
+            var old2NewPath = {},
+                paths = [],
+                relationsToCopyOver = [],
+                copies = [],
+                source, target, oldTarget,
+                pathInDomain = function (path, domain) {
+                    return path === domain | path.indexOf(domain + CONSTANTS.PATH_SEP) === 0;
+                },
+                gatherRelations = function (commonPathInformation, firstNode, secondNode) {
+                    var commonParent,
+                        overlaysToCheck,
+                        i;
 
-            for (i = 0; i < nodes.length; i++) {
-                paths.push(innerCore.getPath(nodes[i]));
+                    commonParent = firstNode;
+                    while (self.getPath(commonParent) !== commonPathInformation.common) {
+                        commonParent = self.getParent(commonParent);
+                    }
+
+                    // first -> second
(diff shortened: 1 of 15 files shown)


src/common/core/corerel.js
```diff
@@ -955,51 +955,180 @@ define([
...
+                                name: overlaysToCheck[i].n
+                            });
+                        }
+                    }
+                },
+                i, j;
+
+            //first we collect the relations that we need to preserve
+            for (i = 0; i < nodes.length; i += 1) {
+                paths.push(self.getPath(nodes[i]));
```
pmeijer: Couldn't this be part of the for loop that follows?

src/common/core/corerel.js
```diff
@@ -955,51 +955,180 @@ define([
...
-            //now we use our simple copy
-            for (i = 0; i < nodes.length; i++) {
-                newNode = self.copyNode(nodes[i], parent, takenRelids, relidLength);
-                copiedNodes.push(newNode);
-                if (takenRelids) {
-                    takenRelids[self.getRelid(newNode)] = true;
-                }
+            //do the actual copying
+            for (i = 0; i < nodes.length; i += 1) {
+                copies.push(self.copyNode(nodes[i], parent));
```
pmeijer: What about the relid reservation?

pmeijer: `takenRelids, relidLength` should be caught by jshint..

src/common/core/coretype.js
```diff
@@ -776,59 +873,112 @@ define([
...
+
+                    oldTarget = self.overlayInquiry(parent, source, relationsToPreserve[i].name);
+                    if (oldTarget !== null && typeof oldTarget.value === 'string') {
+                        self.overlayRemove(parent, source, relationsToPreserve[i].name, oldTarget.value);
+                    }
+                    self.overlayInsert(parent, source, relationsToPreserve[i].name, target);
                 }
             }
 
-            this.processRelidReservation(parent, longestNewRelid);
```
pmeijer: Doesn't this apply anymore?

src/common/core/corerel.js
pmeijer: I see that coretype overrides copyNodes completely...

src/common/core/corerel.js
kecso: that would mean to compute paths O(n^2) times instead of O(n)

pmeijer: nm, I missed the second `paths[j]` access..

src/common/core/users/merge.js
```diff
@@ -24,7 +24,7 @@ define([
 
         if (persisted.hasOwnProperty('objects') === false || Object.keys(persisted.objects).length === 0) {
             parameters.logger.warn('empty patch was inserted - not making commit');
-            return Q({
+            return new Q({
```
pmeijer: There shouldn't be a new prefix here

PR #1362, opened 2017-04-06
Panels can be split arbitrarily many times

### LINKED ISSUE #1361: Allow the user to choose whether to split vertically or horizontally when using the "split view"
And provide a button to swap between the two layout options instead of automatically choosing for them based on the size of the window. :)

pmeijer:
This feature opens up possibilities to write smaller visualizers that end users can include without editing any code. Currently the state of the split panel isn't persisted - but if requested this can be added with option to store the layout for a specific project or project kind.

This PR also cleans the state handling by making sure that only the VisualizerPanel/SplitPanel registers the active panel. (The visualizers themselves should not have to deal with this.)

Closes #1361


finger563:
Feature to remember state of split panel is definitely requested 😃

PR #1364, opened 2017-04-07
Closes #1360 navigate to parent is now owned by tree-browser.

### LINKED ISSUE #1360: Have the "go to parent" button be more conveniently placed
in the modeleditor viz, the `go to parent` button is the last one in the toolbar of buttons, this means that as the window shrinks, it is the first to get pushed into a sub menu.  since it is (in my experience) one of the most commonly used buttons in the toolbar, perhaps make it one of the first buttons instead :)

pmeijer:
This way visualizers do not need to create their own toolbar button for navigating to the parent.


finger563:
I want to upvote this more

PR #1373, opened 2017-04-12
Index appcache manifest and try to avoid browser from caching index.html

pmeijer:
In case index.html is cached - fallback to reload page in start.js if version from index.html doesn't match the actual version.

N.B. Controlling browsers cache mechanism is more or less impossible - therefor the fallback is introduced.

### DIFF
--- a/src/common/util/random.js
+++ b/src/common/util/random.js
@@ -62,6 +62,13 @@ define(['chance', 'common/Constants'], function (ChanceJs, CONSTANTS) {
         return relid;
     }
 
+    function generateRandomString(length) {
+        return chance.string({
+            pool: relidPool,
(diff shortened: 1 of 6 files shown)


src/common/util/random.js
```diff
@@ -62,6 +62,13 @@ define(['chance', 'common/Constants'], function (ChanceJs, CONSTANTS) {
         return relid;
     }
 
+    function generateRandomString(length) {
+        return chance.string({
+            pool: relidPool,
+            length: length
+        });
+    }
+
```
pmeijer: I added this method but ended up not using it.
I left it cause it can be handy (the relidPool) in other use cases.

PR #1378, opened 2017-04-25
Crosscut fixes

### LINKED ISSUE #896: When deleting connectors from a crosscut the object is deleted (not just removed)
Given a connector appearing in more than one crosscut (or the composition).
When that connector is removed from a crosscut it is removed from the model.
It is a bug if there is no way to remove an object from the crosscut without deleting it from the model.
This is true whether the connector was created in the particular crosscut or not.

[crosscut_removal_master.json.txt](https://github.com/webgme/webgme/files/180168/crosscut_removal_master.json.txt)

Copy the 'C' object into the crosscut. Then remove it. It is deleted from the model.
The behavior I expect is that when the last occurrence of an object is removed from
its owner, only then is it be deleted from the model.

### LINKED ISSUE #1266: Crosscut inheritance is not working as it should
Though crosscuts are initially inherited, because multiple information regarding them stored in a single registry entry, whenever an instance changes something that is stored there the inheritance is broken.
- changes that are affecting this: adding new crosscut, renaming a crosscut, removing a crosscut, reordering crosscuts.

After the inheritance broken, though the actual content of the crosscuts still inheriting properly, they cannot be visualized because of the complete overriding of the registry entry.

kecso:
Fixes #896 - removing a connection from a crosscut will only result in the node's deletion, if the given crosscut is the only one it appears and the parent of the connection node is the owner of the crosscut
Fixes #1266 - inheritance among crosscut owners are done via special lookup of the crosscut registry of the owner and merging it along inheritance chain.

The enhancement also contains fixes and improvements on pointer and set member inheritance handling.
A set member is considered relative if it is inside the containment subtree of the root of the instantiation. A relative member inherits relatively meaning it will be 'replaced' by its instance. Otherwise the member considered static and kept as is in the instances.

### DIFF
--- a/src/client/js/Panels/ControllerBase/DiagramDesignerWidgetMultiTabMemberListControllerBase.js
+++ b/src/client/js/Panels/ControllerBase/DiagramDesignerWidgetMultiTabMemberListControllerBase.js
@@ -517,7 +517,7 @@ define(['js/logger',
         return undefined;
     };
 
-    DiagramDesignerWidgetMultiTabMemberListControllerBase.prototype.getMemberListSetsRegistry = function (nodeId) {
+    DiagramDesignerWidgetMultiTabMemberListControllerBase.prototype.getMemberListSetsRegistry = function (/*nodeId*/) {
         this.logger.warn('DiagramDesignerWidgetMultiTabMemberListControllerBase.getMemberListSetsRegistry is not ' +
             'overridden, returning default value...');
         return undefined;
@@ -1600,7 +1600,6 @@ define(['js/logger',
             memberListSetsRegistryKey = this.getMemberListSetsRegistryKey(),
             memberListSetsRegistry,
             i,
-            base,
             len,
             setID;
 
@@ -1640,7 +1639,6 @@ define(['js/logger',
             memberListSetsRegistry,
             i,
             j,
-            base,
             oldIDList = this._tabIDMemberListID,
             urlTab = WebGMEGlobal.State.getActiveTab(),
             setID;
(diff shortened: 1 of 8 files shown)


src/client/js/Panels/ControllerBase/DiagramDesignerWidgetMultiTabMemberListControllerBase.js
```diff
@@ -1598,15 +1604,14 @@ define(['js/logger',
             len,
```
pmeijer: Unused variables..

src/client/js/Panels/ControllerBase/DiagramDesignerWidgetMultiTabMemberListControllerBase.js
```diff
@@ -1640,45 +1645,45 @@ define(['js/logger',
             urlTab = WebGMEGlobal.State.getActiveTab(),
```
pmeijer: same here, check jshint

src/client/js/Panels/Crosscut/CrosscutController.js
```diff
@@ -230,13 +230,13 @@ define(['js/logger',
...
             memberListContainerID = this._memberListContainerID,
-            crosscutsRegistry = GMEConcepts.getCrosscuts(memberListContainerID),
+            crosscutsRegistry = this.getMemberListSetsRegistry(memberListContainerID),
             len = crosscutsRegistry.length;
 
         if (this._autoCreateCrosscut === true && len === 0 && this._tabsHasBeenRequested === false) {
             this._onTabAddClicked();
             memberListContainerID = this._memberListContainerID;
-            crosscutsRegistry = GMEConcepts.getCrosscuts(memberListContainerID);
+            crosscutsRegistry = this.getMemberListSetsRegistry(memberListContainerID);
```
pmeijer: Is `GMEConcepts.getCrosscuts` deprecated now?

PR #1404, opened 2017-05-17
Npm packages and bower updates

pmeijer:
Major version changes: 
superagent `2.2.0 -> 3.5.2`

For minor and micro updates see diffs of `bower.json` and `package.json`.

Not that after this update node < 4 (that is 0.12) is no longer supported.


src/common/storage/socketio/browserclient.js
```diff
@@ -27,7 +27,7 @@ define(['common/util/url'], function (URL) {
                     socket;
 
                 logger.debug('Connecting to "' + hostAddress + '" with options', gmeConfig.socketIO.clientOptions);
-                socket = io.connect(hostAddress, gmeConfig.socketIO.clientOptions);
+                socket = io(hostAddress, gmeConfig.socketIO.clientOptions);
```
pmeijer: this is just to follow the documented approach..

bower.json
```diff
@@ -33,25 +33,24 @@
     "backbone": "1.3.3",
     "underscore": "1.8.3",
     "bootstrap": "3.3.7",
-    "chance": "1.0.3",
+    "chance": "1.0.8",
     "d3": "3.5.16",
-    "visionmedia-debug": "2.2.0",
```
pmeijer: debug no longer comes with a distribution (instead see the UPDATE at `/src/client/lib/debug` same as superagent)

PR #1408, opened 2017-05-18
Ejs based and inmodel SVG

kecso:
This feature enhances the SVG decorator's behavior with the followings:
- the SVG's can be edited and stored textually as part of the model which will make them more portable
- ejs templates that translates to correct SVGs can be used (either in form of file or as an inmodel text in place of SVG registry entries)
- for ejs templates, the clientNode will be provided allowing acces to the attributes and other properties of the given node
- the clientNode API now also provides support for getting nodes (identical to the client API's getNode function) by their ids, but the template make sure to check if the returned value is not null (as it will not load the actual node)
- by using the _hidename_ data tag in your SVG, you can hide the name part of the node's decoration (if your SVG already visualizes it for example) ' data-hidename="true" '

We also added two SVGs to the basic collection to highlight the capabilities of the new feature:
- ChildListerBox makes a simple rounded box that displays the name of the node and the names of its children
- UMLClass visualizes the node and its attributes in a UML fashion


src/client/js/Utils/SvgManager.js
```diff
@@ -0,0 +1,155 @@
...
+        return null;
+    }
+
+    return {
+        getSvgUri: uri,
+        getSvgContent: content,
+        getRawSvgContent: raw,
+        isSvg: isSvg,
+        test: test
+    };
```
pmeijer: Why not keep the same names?
Also these should have a docstring above.

src/client/js/Utils/SvgManager.js
```diff
@@ -0,0 +1,155 @@
...
+        } catch (e) {
+            return e;
+        }
+
+        return null;
+    }
+
+    return {
+        getSvgUri: uri,
+        getSvgContent: content,
```
pmeijer: getSvgElement - returns jquery el
getSvgContent - returns svg string

src/client/js/Dialogs/DecoratorSVGExplorer/DecoratorSVGExplorerDialog.js
```diff
@@ -72,34 +78,150 @@ define(['js/Constants',
...
+        });
+
+        /* jshint ignore:start */
+        this._codemirror = CodeMirror(this._codemirrorEl[0], {
+            readOnly: false,
+            lineNumbers: true,
+            matchBrackets: true,
+            lint: false,
+            theme: 'monokai',
+            mode: 'htmlmixed',
```
pmeijer: Could this work? http://codemirror.net/mode/htmlembedded/

src/client/decorators/SVGDecorator/Core/SVGDecorator.Core.js
```diff
@@ -232,38 +243,37 @@ define([
...
     };
 
-    SVGDecoratorCore.prototype._updateSVGContent = function (svg) {
-        var svgIcon;
-        //set new content
-        this.$svgContent.empty();
-        this.$svgContent.removeClass();
-        this.$svgContent.addClass('svg-content');
-        if (svg) {
-            this.$svgContent.addClass(svg.replace(REGEXP.INVALID_CSS_CHARS, '__'));
```
pmeijer: Make sure to remove this from the docs. I don't think anyone used it..
https://github.com/webgme/webgme/wiki/SVG-Icons

PR #1410, opened 2017-05-24
Update bower.json to resolve paths to Github instead of through bower.io

ghemingway:
Using just the short names for packages in bower.json means that bower must use the bower registry to look up each package name to find the corresponding GitHub repo.  If you add the GitHub username and repo to the path it avoids this step.  We are having problems with a corporate network that blocks bower.io but allows GitHub.


pmeijer:
Thanks Graham, in addition I will suppress the bower install from the postinstall when using the npm-published webgme. That way github shouldn't have to be accessed either (some other networks blocks it but not npm).

ghemingway:
Great.  Thanks so much.

PR #1439, opened 2017-06-26
Closes #1438 captures the missing update on deletion of set members

### LINKED ISSUE #1438: Deleted nodes do not trigger an update event for set owners where they were members
This is for when the actual node is deleted. If the member is removed from the set (and still present in the model) the correct event is triggered.

pmeijer:


src/common/util/jsonPatcher.js
```diff
@@ -741,6 +743,18 @@ define([
                         res.update[absGmePath] = true;
                     }
                 }
+
+                // #1438 This will capture set-owner updates.
```
pmeijer: Not sure if this is the most elegant place for this check..

PR #1448, opened 2017-06-30
Option to disabled the use of user-defined/dynamic SVGs

kecso:
By default the dynamic svg usage is allowed, but as it poses security vulnerability it could be disabled.
This enhancement introduces the option `client.allowUserDefinedSVG`.
When disabled, the user is not allowed to use the SVG text editor and if the svg container registry has some textual content, it will not be run/compiled/rendered. Only relative pathes can be used in this mode.

### DIFF
--- a/src/client/js/Utils/SvgManager.js
+++ b/src/client/js/Utils/SvgManager.js
@@ -24,7 +24,7 @@ define([
     function isSvg(text) {
         var result = false;
 
-        if (DecoratorSVGIconList.indexOf(text) > -1) {
+        if (DecoratorSVGIconList.indexOf(text) > -1 || WebGMEGlobal.gmeConfig.client.allowUserDefinedSVG !== true) {
             return false;
         }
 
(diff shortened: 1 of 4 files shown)


src/client/js/Utils/SvgManager.js
```diff
@@ -24,7 +24,7 @@ define([
     function isSvg(text) {
         var result = false;
 
-        if (DecoratorSVGIconList.indexOf(text) > -1) {
+        if (DecoratorSVGIconList.indexOf(text) > -1 || WebGMEGlobal.gmeConfig.client.allowUserDefinedSVG !== true) {
```
pmeijer: What happens now when there are in-model svgs?

kecso: they will be ignored and the default svg will be used

pmeijer: How can that be? There is no reference to the the default svg here..

kecso: by returning false here
the result of any svg lookup will end up being null that will indicate to use the default

PR #1449, opened 2017-06-30
Improved client source documentation

pmeijer:
Far from complete - but added some sought after docs for methods.

- "Moved" `openLibraryOriginInNewWindow` to Library Manager (still available at client).
- Allow passing of empty string to transaction message.

### DIFF
--- a/src/client/js/client/pluginmanager.js
+++ b/src/client/js/client/pluginmanager.js
@@ -30,7 +30,7 @@ define([
         var self = this,
             logger = mainLogger.fork('PluginManager');
 
-        this.getCurrentPluginContext = function (pluginId, activeNodeId) {
+        this.getCurrentPluginContext = function (pluginId, activeNodeId, activeSelectionIds) {
             var activeNode,
                 validPlugins,
                 context =  {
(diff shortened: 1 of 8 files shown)


src/docs/client.doc.js
```diff
@@ -25,4 +27,467 @@
...
+ * @description Remove the library from the current project and branch.
+ * @function removeLibrary
+ * @memberOf Client
+ * @instance
+ * @param {string} libraryName - Name of library to remove.
+ */
+
+/**
+ * @description Rename the library in the current project and branch.
+ * @function removeLibrary
```
kecso: renameLibrary

src/client/js/client/pluginmanager.js
```diff
@@ -77,7 +81,7 @@ define([
          * @param {string} pluginId - id of plugin.
```
kecso: Isn't this description a redundant one now that the doc file also has it?

src/client/js/client/pluginmanager.js
```diff
@@ -107,7 +111,7 @@ define([
...
```
kecso: Same redundancy question goes for this as well

src/docs/client.doc.js
pmeijer: Good catch!

src/client/js/client/pluginmanager.js
pmeijer: Let's keep the inlined docs. That file is short as it is and can be useful too.

PR #1470, opened 2017-07-24
Plugin API for invoking other plugins from a plugin

pmeijer:
Dependency plugins can be specified in the metadata with the option to have plugin configurations propagate to the UI.

Invoked plugins can be configured to run under a name space relative the invoker.

Calls to save a bypassed in a transaction like manner and only the outer most plugin have the option to save. (This is to avoid potential inconsistencies in the shared core nodes and it also gives more control to the invoker).

https://github.com/webgme/webgme/wiki/GME-Plugins#invoking-other-plugins

### DIFF
--- a/src/plugin/PluginBase.js
+++ b/src/plugin/PluginBase.js
@@ -662,10 +662,22 @@ define([
     };
 
     /**
+     * Initializes and invokes the given plugin (at pluginId).
+     * Things to note:
+     *  1. If the invoked plugin calls save - it will not persist nor make a commit. The message will be recorded in
+     *  the InterPluginResult.
+     *  2. Artifacts and files saved will be added to the blob-storage. Invoked plugins can expose the content by adding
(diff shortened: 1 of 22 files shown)


src/plugin/PluginBase.js
```diff
@@ -438,10 +451,20 @@ define([
 
         commitMessage = message ? commitMessage + ' - ' + message : commitMessage;
 
+        if (this.callDepth > 0) {
+            self.logger.debug('Call-depth is greater than zero, will not persist "', this.callDepth, '"');
+            // How to handle this neatly..
+            self.addCommitToResult(commitMessage);
```
kecso: It is probably another function like 'addCommitMessage'.

pmeijer: It would have worked, but I changed it to call `self.result.addCommitMessage` right away

Issue #1451, opened 2017-07-05
A codemirror popup for editing multi-line attributes in the property editor would be useful


DWatkins:
It would be very convenient to allow specification in the meta that an attribute is multi-line, and if so, to open a code mirror similar to when editing constraints.

pmeijer:
Closed by #1485

PR #1485, opened 2017-08-16
Meta attribute enhancements

### LINKED ISSUE #1480: Read-only mode is not triggered properly
When selecting a given commit, the read-only mode is not shown on the part-browser and the property-editor.

### LINKED ISSUE #1482: Meta-readonly attributes
It would be beneficial to have a meta-readonly flag for the attribute definitions.
That way only the language components could change those attributes, but not the actual models.

### LINKED ISSUE #1483: Multi-line string attributes
As the number of cases when multi-line strings are captured in attributes, it would be beneficial to support that at the point of definition.
Furthermore, WebGME could provide several sub-types for attributes (like javascript, or java, or modelica code) that can control how the properties could  be edited in the property editor.
The property editor - by leveraging the codemirror library - could provide a more suitable way of editing these attributes.

kecso:
Closes #1483, #1451 Multi-line string attributes:
- during the meta-attribute definition, the user can choose to set the string as `multiline` and further set a `subtype` (c-like, java, etc...). The property-editor will interpret this by adding a button/link that will initiate a pop-up, codemirror based editing dialog. The sub-type is the basis of the highlighting set for the editor.

Closes #1482 Meta-readonly attributes:
- the user can set for any attribute definition, to be 'meta-readonly', that means that only nodes that are part of the meta can change the value
- the functionality is a Core concept, so even plugins will not be able to change such attribute values for non-meta elements

Closes #1482 Propagate meta-rule removal:
- as a continuation of the meta concept/rule renaming, the WebGME now offers to propagate a rule removal throughout the whole project
- this can make life easier as the 'left-over' values removed in a single step

Fixes #1480 Read-only mode is not triggered properly

### DIFF
--- a/src/client/decorators/MetaDecorator/DiagramDesigner/AttributeDetailsDialog.js
+++ b/src/client/decorators/MetaDecorator/DiagramDesigner/AttributeDetailsDialog.js
@@ -8,9 +8,9 @@ define([
     'js/util',
     'common/regexp',
     'text!./templates/AttributeDetailsDialog.html',
-    'js/Dialogs/CodeEditor/constants',
+    'common/Constants',
     'css!./styles/AttributeDetailsDialog.css'
-], function (util, REGEXP, attributeDetailsDialogTemplate, MODE_CONSTANTS) {
+], function (util, REGEXP, attributeDetailsDialogTemplate, CONSTANTS) {
 
     'use strict';
 
@@ -98,7 +98,7 @@ define([
                 self._pRegExp.hide();
                 self._pRegExpValue.val('');
             } else {
-                self._multilineType.val(MODE_CONSTANTS.MODE.generic);
+                self._multilineType.val(CONSTANTS.MULTILINE_MODE.generic);
                 self._pMultiLineSubTypes.hide();
                 self._pRegExp.show();
             }
@@ -113,13 +113,13 @@ define([
                     type: self._inputType.val(),
                     defaultValue: self._inputDefaultValue.val(),
                     isEnum: self._cbEnum.is(':checked'),
-                    metaReadOnly: self._cbMetaReadOnly.is(':checked'),
+                    readonly: self._cbReadonly.is(':checked'),
                 },
                 cValue;
 
             if (self._cbMultiLine.is(':checked')) {
                 attrDesc.multiline = true;
-                attrDesc.multilineType = MODE_CONSTANTS.MODE[self._multilineType.val()] || MODE_CONSTANTS.MODE.generic;
+                attrDesc.multilineType = self._multilineType.val() || CONSTANTS.MULTILINE_MODE.generic;
             }
 
             if (attrDesc.isEnum) {
@@ -266,10 +266,10 @@ define([
         this._el = this._dialog.find('.modal-body').first();
         this._cbEnum = this._el.find('#cbEnum').first();
         this._pEnum = this._el.find('#pEnum').first();
-        this._cbMultiLine = this._el.find('#cbMultiLine').first();
+        this._cbMultiLine = this._el.find('#cbMultiline').first();
 
-        this._pMultiLine = this._el.find('#pMultiLine').first();
-        this._pMultiLineSubTypes = this._el.find('#pMultiLineSubTypes').first();
+        this._pMultiLine = this._el.find('#pMultiline').first();
+        this._pMultiLineSubTypes = this._el.find('#pMultilineSubtypes').first();
         this._pMultiLineSubTypes.hide();
         this._multilineType = this._el.find('#multilineType').first();
 
@@ -289,7 +289,7 @@ define([
 
         this._inputEnumValues = this._el.find('#inputEnumValues').first();
 
-        this._cbMetaReadOnly = this._el.find('#cbMetaReadOnly').first();
+        this._cbReadonly = this._el.find('#cbReadonly').first();
 
         //extended options
         this._pRegExp = this._el.find('#pRegExp');
@@ -383,8 +383,8 @@ define([
         //fill controls based on the currently edited attribute
         this._inputName.val(attributeDesc.name);
         this._inputType.val(attributeDesc.type);
-        if (attributeDesc.metaReadOnly) {
-            this._cbMetaReadOnly.attr('checked', true);
+        if (attributeDesc.readonly) {
+            this._cbReadonly.attr('checked', true);
         }
         if (attributeDesc.type === 'boolean') {
             //boolean type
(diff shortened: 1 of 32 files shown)


src/client/decorators/MetaDecorator/DiagramDesigner/AttributeDetailsDialog.js
```diff
@@ -8,8 +8,9 @@ define([
     'js/util',
     'common/regexp',
     'text!./templates/AttributeDetailsDialog.html',
+    'js/Dialogs/CodeEditor/constants',
     'css!./styles/AttributeDetailsDialog.css'
-], function (util, REGEXP, attributeDetailsDialogTemplate) {
+], function (util, REGEXP, attributeDetailsDialogTemplate, MODE_CONSTANTS) {
```
pmeijer: The MetaDecorator should use a enumeration from `common/Constants.js`, the js/Dialogs/CodeEditor/constants should provide a mapping from that to the code-editor language.

src/client/js/Controls/PropertyGrid/PropertyGridWidgets.js
```diff
@@ -15,6 +15,7 @@ define([], function () {
         META_TYPE_WIDGET: 'META_TYPE_WIDGET',
         MULTI_SELECT_WIDGET: 'MULTI_SELECT_WIDGET',
         COLOR_PICKER: 'COLOR_PICKER',
-        SVG_SELECT: 'SVG_SELECT'
+        SVG_SELECT: 'SVG_SELECT',
+        MULTI_LINE: 'MULTI_LINE'
```
pmeijer: MULTILINE

src/client/js/Dialogs/CodeEditor/constants.js
```diff
@@ -0,0 +1,31 @@
...
+            JavaScript: 'javascript',
+            Markdown: 'markdown',
+            Mathematica: 'mathematica',
+            Modelica: 'modelica',
+            LaTeX: 'stex',
+            Lua: 'lua',
+            SQL: 'sql',
+            Python: 'python',
+            TTCN: 'ttcn',
+            YAML: 'yaml'
```
pmeijer: The LHS should be defined in `common/Constants.js` and used here.

src/client/js/Panels/MetaEditor/MetaEditorControl.js
```diff
@@ -1240,11 +1242,39 @@ define(['js/logger',
...
-            objectNode = this._client.getNode(objectID);
+        var self = this,
+            containerNode = this._client.getNode(containerID),
+            objectNode = this._client.getNode(objectID),
+            confirmDialog = new ConfirmDialog();
 
         if (containerNode && objectNode) {
-            this._client.delChildMeta(containerID, objectID);
+            confirmDialog.show({
+                title: 'Propagate Meta containment remoe',
```
pmeijer: remoe

src/common/Constants.js
```diff
@@ -86,6 +85,26 @@ define(['common/core/constants', 'common/storage/constants'], function (CORE, ST
             RENAME_META_POINTER_TARGET: 'renameMetaPointerTarget',
             CHANGE_ASPECT_META: 'changeAspectMeta',
             REMOVE_META_RULE: 'removeMetaRule'
+        },
+
+        MULTILINE_MODE: {
```
pmeijer: `ATTRIBUTE_MULTILINE_TYPES`
Also extend the core-docs with examples of the new "properties" `getAttributeMeta`

src/common/Constants.js
```diff
@@ -86,6 +85,26 @@ define(['common/core/constants', 'common/storage/constants'], function (CORE, ST
...
+            generic: 'generic',
```
pmeijer: Doesn't `plaintext` fit better than `generic`?

src/common/Constants.js
```diff
@@ -86,6 +85,26 @@ define(['common/core/constants', 'common/storage/constants'], function (CORE, ST
...
+            JavaScript: 'JavaScript',
+            Markdown: 'Markdown',
+            Mathematica: 'Mathematica',
+            Modelica: 'Modelica',
+            LaTeX: 'LaTeX',
+            Lua: 'Lua',
+            SQL: 'SQL',
+            Python: 'Python',
+            TTCN: 'TTCN',
+            YAML: 'YAML'
```
pmeijer: I would have all of these in all lower-cases. Also breakdown `C-LIKE` to `c` and `cpp` (they can still map to codemirrors c-like). If `csharp` is highlighted in `clike` too..

src/client/js/Dialogs/CodeEditor/CodeEditorDialog.js
```diff
@@ -0,0 +1,116 @@
...
+            codemirrorOptions = {
+                readOnly: params.readOnly,
+                lineNumbers: true,
+                matchBrackets: true,
+                fullscreen: false,
+            },
+            oked = false;
+
+        if (params.multilineType && params.multilineType !== COMMON.MULTILINE_MODE.generic) {
+            codemirrorOptions.mode = CONSTANTS.MODE[params.multilineType];
```
pmeijer: Can the mode be set by the user as well? Would be good with a dropdown list where the mode from the meta is selected by default. That way it's also clear which type of code is expected to be typed.

src/client/decorators/MetaDecorator/DiagramDesigner/AttributeDetailsDialog.js
```diff
@@ -119,7 +119,7 @@ define([
 
             if (self._cbMultiLine.is(':checked')) {
                 attrDesc.multiline = true;
-                attrDesc.multilineType = self._multilineType.val() || CONSTANTS.MULTILINE_MODE.generic;
+                attrDesc.multilineType = self._multilineType.val() || CONSTANTS.ATTRIBUTE_MULTILINE_TYPES.plain;
```
pmeijer: This should be `plaintext` as well

src/common/Constants.js
```diff
@@ -87,23 +87,26 @@ define(['common/core/constants', 'common/storage/constants'], function (CORE, ST
...
-            SQL: 'SQL',
-            Python: 'Python',
-            TTCN: 'TTCN',
-            YAML: 'YAML'
+        ATTRIBUTE_MULTILINE_TYPES: {
+            plaintext: 'plaintext',
+            html: 'html',
+            c: 'c',
+            cpp: 'cpp',
+            'c#': 'c#',
```
pmeijer: I would avoid special characters in these type of enumerations.

src/common/core/users/metarules.js
```diff
@@ -272,6 +272,13 @@ define(['q', 'common/core/constants'], function (Q, CONSTANTS) {
 
         for (i = 0; i < names.length; i++) {
             if (validNames.indexOf(names[i]) !== -1) {
+                if (meta.attributes[names[i]].readonly === true && core.getOwnAttribute(node, names[i]) !== undefined &&
+                    core.isMetaNode(node) === false) {
+                    result.messages.push('Read-only attribute "' + names[i] +
+                        '" value has been set for a non-meta node!');
+                    result.hasViolation = true;
+                }
```
pmeijer: No test for this?

PR #1488, opened 2017-08-18
Maximize split panel

kecso:
The feature provides a pop-up button to maximize a split panel and later to return to the previous split configuration.


pmeijer:
Create a new branch from the old one and once it's fixed, make a new PR

PR #1497, opened 2017-08-24
Fixes #1489 Non-meta nodes in aspects throws exception in getValidChildrenMetaNodes

### LINKED ISSUE #1489: Non-meta nodes in aspects throws exception in getValidChildrenMetaNodes
It should be noted that a meta like that is reported as inconsistent.

The same issue most likely holds for containment and sets members too.

```
Uncaught TypeError: Cannot read property 'relid' of undefined
    at MetaCore.CoreTree.getPath (:8888/common/core/coretree.js:391)
    at sameNode (:8888/common/core/metacore.js:33)
    at MetaCacheCore.MetaCore.isTypeOf (:8888/common/core/metacore.js:138)
    at isTypeOf (:8888/common/core/mixincore.js:313)
    at MixinCore.isTypeOf (:8888/common/core/mixincore.js:390)
    at Object.MetaQueryCore.getValidChildrenMetaNodes (:8888/common/core/metaquerycore.js:148)
    at CoreQ.Core.getValidChildrenMetaNodes (:8888/common/core/core.js:3207)
    at GMENode.getValidChildrenTypesDetailed (:8888/js/client/gmeNodeGetter.js:361)
    at PartBrowserControl._getPartDescriptorCollection (:8888/panel/PartBrowser/PartBrowserPanelControl.js:302)
    at :8888/panel/PartBrowser/PartBrowserPanelControl.js:81
```

kecso:
As out-of-meta nodes can be part of meta rules, these specific inquiry functions should be prepared for that scenario.
The function results remains the same, but it will work only with the paths of the rule related nodes instead of trying to get them from all the meta nodes.

### DIFF
--- a/src/client/js/client/gmeNodeSetter.js
+++ b/src/client/js/client/gmeNodeSetter.js
@@ -1001,7 +1001,7 @@ define([], function () {
         function isTypeOf(path, typePath) {
             var node = _getNode(path);
 
-            if (node && typeNode) {
+            if (node) {
                 return state.core.isTypeOf(node, typePath);
             }
 
--- a/src/client/js/client/gmeNodeSetter.js
+++ b/src/client/js/client/gmeNodeSetter.js
@@ -999,11 +999,10 @@ define([], function () {
         // TODO: These should be removed at next version bump.
 
         function isTypeOf(path, typePath) {
-            var node = _getNode(path),
-                typeNode = _getNode(typePath);
+            var node = _getNode(path);
 
             if (node && typeNode) {
-                return state.core.isTypeOf(node, typeNode);
+                return state.core.isTypeOf(node, typePath);
             }
 
             return false;
--- a/src/common/core/metaquerycore.js
+++ b/src/common/core/metaquerycore.js
@@ -98,17 +98,6 @@ define([
                 delete rules.max;
                 delete rules.min;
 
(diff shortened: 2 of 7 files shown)


src/common/core/metaquerycore.js
```diff
@@ -98,21 +98,10 @@ define([
...
-                    } else {
-                        delete rules[keys[i]];
-                    }
-                }
-
                 keys = Object.keys(rules);
                 for (i = 0; i < children.length; i += 1) {
                     for (j = 0; j < keys.length; j += 1) {
-                        if (innerCore.isTypeOf(children[i], metaNodes[keys[j]])) {
+                        if (innerCore.isTypeOfPath(children[i], keys[j])) {
```
pmeijer: Can't this method be exposed on the core? Or alternatively `isTypeOf` could accept both a node and path as the second argument. What definitely should be done is to merge the implementation of these two.

src/client/js/client/gmeNodeSetter.js
```diff
@@ -999,11 +999,10 @@ define([], function () {
         // TODO: These should be removed at next version bump.
 
         function isTypeOf(path, typePath) {
-            var node = _getNode(path),
-                typeNode = _getNode(typePath);
+            var node = _getNode(path);
```
pmeijer: This method will throw an undefined error

PR #1543, opened 2017-11-02
Fixes #1542 Document Editor Dialog Title can be changed with optional constructor argument

### LINKED ISSUE #1542: Allow DocumentEditorDialog title to be changed through its initialize function
It's a nice reusable editor/viewer for markdown, would be nice to be able to configure the title to say something other than `Edit Comment Documentation`

finger563:


src/client/decorators/DocumentDecorator/DiagramDesigner/DocumentEditorDialog.js
```diff
@@ -28,6 +28,9 @@ define(['js/util',
         this._el = this._dialog.find('.modal-body').first();
         this._btnSave = this._dialog.find('.btn-save').first();
         this._pMeta = this._el.find('#pMeta').first();
+        this._title = this._dialog.find('#dialog-title').first();
+        if (title)
```
pmeijer: Please include curly brackets around block statements.

src/client/decorators/DocumentDecorator/DiagramDesigner/DocumentEditorDialog.html
```diff
@@ -3,7 +3,7 @@
         <div class="modal-content">
             <div class="modal-header">
                 <button type="button" class="close" data-dismiss="modal">×</button>
-                <h3>Edit Comment Documentation</h3>
+                <h3 id="dialog-title">Edit Comment Documentation</h3>
```
pmeijer: In general I would avoid using id attributes, but in this case since it's a modal it's not very probably that multiple ones are displayed at the same time. (And four lines down there's already an id attribute.

finger563: So... do you want me to change it or not? :)

pmeijer: You can leave it as it’s consistent with the other element and changing it might require changing some styles rules.

PR #1550, opened 2017-11-15
Create source map for client/dist/webgme....lib.build.js

ksmyth:
It's useful for debugging plugins to have the source map for the dist/lib...js file.


pmeijer:
I would recommend using the `/debug.html` route for debugging.

ksmyth:
The source map worked when I used it.

I did not know about `/debug.html`. That is helpful, thanks

PR #1575, opened 2018-02-08
Closes #1582 code-editor passes watcherId to doc s.t. multiple conns can be handled

pmeijer:
This fix will take effect once the engine has been bumped to include [webgme-engine#47](https://github.com/webgme/webgme-engine/pull/47).


finger563:
Yay! 👍

PR #1580, opened 2018-03-08
Add and expose sortable widget.

finger563:
Still in the progress of implementing this.


finger563:
I've tested and it works as intended - the resultant array is exposed through the `propertyValue` of the widget, just like every other widget.

finger563:
Usage: provide `sortable` as your `propertyType` for your widget config (e.g. in your plugin config widget or metadata) and provide values in your `valueItems`.

pmeijer:
Thanks will, appreciate the PR 👍

PR #1581, opened 2018-03-12
Add description and hidden in schema for Attributes

kecso:
With this enhancement, the user is now able to add a description to the Attribute definition which become visible as the tooltip text if the modeler hovers over the name in the property editor.
Also, we introduced the hidden option for attributes, that will be used also in the property editor by not showing the attribute (but it can be manipulated via programmatic API's )

### DIFF
--- a/src/client/decorators/MetaDecorator/DiagramDesigner/templates/AttributeDetailsDialog.html
+++ b/src/client/decorators/MetaDecorator/DiagramDesigner/templates/AttributeDetailsDialog.html
@@ -17,7 +17,7 @@
                     <label class="col-sm-4 control-label" for="description">Description</label>
 
                     <div class="col-sm-8 controls">
-                        <textarea id="description" placeholder="What is this attribute for" rows="3" style="width:80%;"/>
+                        <textarea id="description" placeholder="Enter an optional description.." rows="3" style="width:80%;"/>
                     </div>
                 </div>
                 <div class="form-group">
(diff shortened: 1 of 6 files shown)


src/client/decorators/MetaDecorator/DiagramDesigner/templates/AttributeDetailsDialog.html
```diff
@@ -13,6 +13,13 @@ <h3>Edit attribute parameters...</h3>
                         <input type="text" id="inputName" placeholder="Name">
                     </div>
                 </div>
+                <div class="form-group" id="pDescription">
+                    <label class="col-sm-4 control-label" for="description">Description</label>
+
+                    <div class="col-sm-8 controls">
+                        <textarea id="description" placeholder="What is this attribute for" rows="3" style="width:80%;"/>
```
pmeijer: `What is this attribute for` -> `Enter an optional description..`

src/client/decorators/MetaDecorator/DiagramDesigner/templates/AttributeDetailsDialog.html
```diff
@@ -34,6 +41,12 @@ <h3>Edit attribute parameters...</h3>
                             Read-only
                         </label>
                     </div>
+                    <div id="pHidden" class="col-sm-2 controls" title="If true, value will not be visible in the property editor">
```
pmeijer: `If true, value will not be visible in the property editor` -> `If true, the attribute will not be displayed in the GUI`

PR #1593, opened 2018-04-09
Closes #1591 Show info in project browser and use display name for users if defined

### LINKED ISSUE #1591: Show info in project browser
It would be beneficial to show (and potentially allow editing) the info-metadata of the projects when they are listed in the project list dialog.

### LINKED ISSUE #1592: Allow inferred users to have displayName
Other systems might use non-readable userIds and they might allow users to pick their visible name (or simply use their full name). Therefore it would be nice if webgme could have an additional field among the user data that could store this info.
This also means that ui pieces should replace the userid with this displayName if that is available.

kecso:
Closes #1592 Allow inferred users to have displayName.

Users having displayName will now use it in every UI element that provides the user info (user badge, multiline property edit, project management).
Also as a side-effect, the display name of a project is changed. If the project is owned by the user, its display name will only be the project name. 'owner / projectName' remains for all other projects.


src/client/js/Dialogs/CodeEditor/CodeEditorDialog.js
```diff
@@ -377,7 +377,7 @@ define([
                         if (otherClients.hasOwnProperty(eData.socketId) === false) {
                             colorIndex = Object.keys(otherClients).length % CLIENT_COLORS.length;
                             otherClients[eData.socketId] = {
-                                userId: eData.userId,
+                                userId: WebGMEGlobal.getDisplayName(eData.userId),
```
pmeijer: The `getDisplayName` is ambiguous when directly on `WebGMEGlobal`.

Issue #1553, opened 2017-11-22
feature request: Allow plugin config to have a nested structure for collapsible headers


DWatkins:
When a plugin has many config options to set the UI can get crowded at times.  Having a nested structure for the config would allow the use of collapsible headers, e.g. `Advanced Options`

finger563:
Perhaps just an addition that allows the user to specify the template for the plugin as part of the call to `PluginConfigDialog.prototype.show`?

PR #1621, opened 2018-07-21
Add nesting capability to Plugin Config Dialog

finger563:
Nesting automatically uses `<details>` to enable nesting in the UI and allows config keys to store objects which themselves contain config keys. For #1553

### DIFF
--- a/src/client/js/Dialogs/PluginConfig/PluginConfigDialog.js
+++ b/src/client/js/Dialogs/PluginConfig/PluginConfigDialog.js
@@ -215,10 +215,29 @@ define([
 
                 el.find('summary.control-label').text(pluginConfigEntry.displayName);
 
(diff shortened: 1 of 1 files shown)


finger563:
This is incomplete right now - need to enable arbitrary nesting (instead of just one level like it is now)

finger563:
Have implemented nesting of headers - need to update the styling now.

finger563:
As a note - just saw that webgme already has bootstrap - so I'll try to move the them over to be `collapse` buttons and divs.

finger563:
I've updated the styling to a point where I think it's not too bad and isn't too out of line with what webgme already has.  I think this PR is ready for some review :)

finger563:
Thanks!

PR #1679, opened 2019-04-19
Navigation title

### LINKED ISSUE #1677: Navigation title
Navigation up on the containment hierarchy can be hard with only the 'up' button and the breadcrumbs header.
It was requested that the title on the 'ModelEditor' might become a clickable navigation panel (configure the number of upper levels that are offered and the attribute it lists - by default it should be the name as currently).

kecso:
As an additional way to quickly navigate towards the root node of the project, we added the Navigation Title feature.
If allowed (currently only in the Composition visualizer, but can be easily adapted to any visualizer using the PanelBaseWithHeader) the title will be a clickable breadcrumbs line (navigation reacts to double click!) instead of a plain, static label.
The feature has to be configured with components config.
The following is an example on how to allow for every project on a given deployment:
```
"GenericUIModelEditorPanel": {
    "navigationTitle": {
      "enabled": true,
      "attribute": "name",
      "depth": 4
    },
    "byProjectKind": {
      "navigationTitle": {}
    },
    "byProjectId": {
      "navigationTitle": {}
    }
  }
```

About the options:
- if attribute is not given, the id of the node will be shown on the screen
- the depth can be any number and controlls how many container (starting from the root), will be visualized and clickable for a quick navigation

This resolves #1677.

### DIFF
--- a/src/client/js/PanelBase/PanelBaseWithHeader.js
+++ b/src/client/js/PanelBase/PanelBaseWithHeader.js
@@ -79,6 +79,9 @@ define([
         this.$panelHeader = $('<div/>', {
             class: 'panel-header no-print'
         });
+        this.$panelHeaderTitle = $('<div/>', {
+            class: 'panel-header-title'
+        });
 
         //Create Panel's BODY
         //set $el to panel-body for subclass use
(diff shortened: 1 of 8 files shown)


src/client/js/PanelBase/PanelBaseWithHeader.js
```diff
@@ -91,37 +98,154 @@ define(['js/PanelBase/PanelBase', 'css!./styles/PanelBaseWithHeader.css'], funct
...
+    PanelBaseWithHeader.prototype._shouldUseNavigationTitle = function () {
+        var config = {enabled: false};
+        if (this._config && this._client) {
+            config = this._getCurrentNavigationConfig();
+        }
+        return config.enabled === true;
+    };
+
+    PanelBaseWithHeader.prototype._refreshNavigationTitle = function () {
+        console.log(WebGMEGlobal.State.getActiveObject());
```
kecso: remove

src/client/js/PanelBase/PanelBaseWithHeader.js
```diff
@@ -91,37 +98,154 @@ define(['js/PanelBase/PanelBase', 'css!./styles/PanelBaseWithHeader.css'], funct
...
     };
 
     PanelBaseWithHeader.prototype.destroy = function () {
+        this._detachClientEventListener();
         this.clear();
         this.$_el.remove();
         this._destroyedInstance = true;
     };
 
+    PanelBaseWithHeader.prototype._configNavigationTitle = function (config) {
```
kecso: is it really used?

PR #1690, opened 2019-12-16
Password - attribute

kecso:
Now the user is able to set a string attribute as password.
This means that by default the UI will hide the actual value (using the 'password' type of input fields), which can then be turned on and off by the additional eye icon in the property editor.
This is the only safeguard this option entails, the value itself is not completely protected, whoever has access to the project, can export it, and use it (or change the meta and see the value...)


brollb:
Overall, it looks like a good feature! I noticed a little bug when overwriting the default value for a password field:
DeepinScreenshot_select-area_20200127143330

PR #1749, opened 2022-06-07
Add support for node creation, deletion, rename in GraphViz. Close #1748

### LINKED ISSUE #1748: Edit models from the GraphViz visualizer
Currently, the GraphViz visualizer is read-only. It would be good if we could add, remove, and rename nodes from there.

brollb:
This is mostly ready but needs a better way for the user to input the node name on rename. Unfortunately `$.editInPlace` doesn't work for this case since the text box loses focus if the cursor ever leaves the element's bounding box. I am not exactly sure of the cause; I haven't seen it before with `editInPlace` nor `d3`. That said, I am not really convinced it is even the best UX or really required. Currently, I am thinking something like [this](https://code.daypilot.org/17463/javascript-prompt-replacement) might be a better fit anyway.

Here is a screenshot of the addition:
DeepinScreenshot_select-area_20220607163055

### DIFF
--- a/src/client/js/Panels/GraphViz/GraphVizPanelControl.js
+++ b/src/client/js/Panels/GraphViz/GraphVizPanelControl.js
@@ -68,10 +68,6 @@ define(['js/logger',
             this._client.deleteNode(nodeId);
         };
 
-        this._graphVizWidget.setName = (nodeId, name) => {
-            this._client.setAttribute(nodeId, 'name', name);
-        };
-
         this._graphVizWidget.onNodeClose = function (id) {
             var deleteRecursive,
                 node,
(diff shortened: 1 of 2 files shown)


kecso:
I would leave out the rename to begin with.
On the other hand, that pop-up stuff looks way better and seems more controlled in terms of user input flow.
It would be sooo nice to have a configurable context menu feature. You create plugins with specific pop-up config and function and offer it as a menu item. The combinations would be endless and the variability of the user interface as well...

All in all, I would leave out the rename for now.

brollb:
Sure, I can leave it out of this PR. The motivation for adding it is to be able to edit a taxonomy (since the name and node types are all that need to be edited) entirely within the GraphViz visualizer.

kecso (approved):

PR #1761, opened 2022-11-02
Add Option for BranchStatusWidget to automatically follow remote at concurrent edits

pmeijer:
With the settings enabled each time a concurrent branch update happens - the user coming in late automatically drops his/her last change and continues on from the other users change(s). 

This will only happen when there's a single commit in the queue - if there are more it falls back to the standard behavior of requiring the user to select action.

To enable feature modify the components-settings with:
```
{
 ...
  "GenericUIBranchStatus": {
    "autoFollowBranchWhenOutOfSync": true
  },
...
```

### DIFF
--- a/src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
+++ b/src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
@@ -27,7 +27,7 @@ define([
 
     function _getDefaultConfig() {
         return {
-            autoFollowBrancWhenOutOfSync: true
+            autoFollowBrancWhenOutOfSync: false
         };
     }
 
--- a/src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
+++ b/src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
@@ -27,7 +27,7 @@ define([
 
     function _getDefaultConfig() {
         return {
-            autoFollowBrancWhenOutOfSync: false
+            autoFollowBranchWhenOutOfSync: false
         };
     }
 
@@ -233,7 +233,7 @@ define([
         this._ddBranchStatus.clear();
         this._outOfSync = true;
         if (this._client.isConnected()) {
-            if (this._config.autoFollowBrancWhenOutOfSync && this._client.getCommitQueue().length === 1) {
+            if (this._config.autoFollowBranchWhenOutOfSync && this._client.getCommitQueue().length === 1) {
                 $.notify({
                     icon: 'fa fa-share-alt fa-rotate-90',
                     message: 'Ooops! Looks like someone is chaning the model concurretly .. ' +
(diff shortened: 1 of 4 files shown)


src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
```diff
@@ -218,6 +233,26 @@ define([
         this._ddBranchStatus.clear();
         this._outOfSync = true;
         if (this._client.isConnected()) {
+            if (this._config.autoFollowBranchWhenOutOfSync && this._client.getCommitQueue().length === 1) {
+                $.notify({
+                    icon: 'fa fa-share-alt fa-rotate-90',
+                    message: 'Ooops! Looks like someone is chaning the model concurretly .. ' +
+                    'Your last change was dropped, check branch history for details.'
```
kecso, line 240: Don't think the history is going to tell anything about the dropped changes, which would be the meaningful info in this case (so I would just let the user know that the change was dropped).

src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
```diff
@@ -218,6 +233,26 @@ define([
         this._ddBranchStatus.clear();
         this._outOfSync = true;
         if (this._client.isConnected()) {
+            if (this._config.autoFollowBranchWhenOutOfSync && this._client.getCommitQueue().length === 1) {
```
kecso, line 236: If we allow users to go this direction, probably we can let them go full on, meaning to drop all changes. I would keep the single drop as an option, but offer to drop no matter the amount of changes lost...

kecso (commented):
Nicely done.
See my comments, but in general it is good to go.

src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
pmeijer: The commit will be head-less, but still available if you check `Recent commits`.
(Not super useful, but just in case someone wrote a novel and hit save xD)

src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
pmeijer: I don't agree. 
1) The first "out of sync" commit is persisted (when the queue builds up they proceeding ones aren't).
2) With only one commit it should be clear what was dropped (in most cases you'd get direct feedback since you're in that context). Knowing that only one commit can be dropped ensures that the user doesn't have to track back what was dropped.

kecso: My bad.
Yeah, forgot that here only the branch update fails, but the commit goes through.
I still wonder though if the scenario could even happen and if we should at some point allow scratching that...

src/client/js/Widgets/BranchStatus/BranchStatusWidget.js
kecso: True.

PR #1788, opened 2023-11-04
Add some modules/methods on webgme server module of type-defs

pmeijer:
This is still very much WIP, but might as well merge it in now.

### DIFF
--- a/typings/webgme.d.ts
+++ b/typings/webgme.d.ts
@@ -1968,8 +1968,8 @@ declare namespace GmeClasses {
          * @return 
          */
         loadSubTree: {
-            (node: Core.Node, callback: GmeCommon.ResultCallback<Core.Node>): void;
-            (node: Core.Node): Promise<Core.Node>;
+            (node: Core.Node, callback: GmeCommon.ResultCallback<Core.Node[]>): void;
+            (node: Core.Node): Promise<Core.Node[]>;
         }
         /**
          * TODO
--- a/typings/webgme.d.ts
+++ b/typings/webgme.d.ts
@@ -138,7 +138,7 @@ declare module "webgme" {
         /** Ensures the user is authenticated. */
         ensureAuthenticated: Function;
         /** If authenticated retrieves the userId from the request. */
-        getUserId: (req: Request) => string;
+        getUserId: (req: any) => string;
         /** Authorization module. */
         gmeAuth: GmeAuth;
         /** Accesses the storage and emits events (PROJECT_CREATED, COMMIT..). */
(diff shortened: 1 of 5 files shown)


typings/webgme.d.ts
```diff
@@ -1574,7 +1681,7 @@ declare namespace GmeClasses {
          * @return a dictionary where every the key of every entry is an absolute path of a set owner node. 
          * The value of each entry is an array with the set names in which the node can be found as a member.
          */
-        isMemberOf(node: Core.Node): Core.DataObject;
```
brollb, line 1577: What was Core.DataObject again?

typings/webgme.d.ts
```diff
@@ -1574,7 +1681,7 @@ declare namespace GmeClasses {
...
+        isMemberOf(node: Core.Node): object;
```
brollb, line 1684: I would use a better return value. Perhaps something like
```
type MemberSets = {[ownerPath: string]: string[]};
```
Not great but better than an opaque object.

typings/webgme.d.ts
```diff
@@ -2488,10 +2595,10 @@ declare namespace GmeClasses {
          */
         makeCommit: {
             (branchName: string, parents: GmeStorage.CommitHash[],
-                rootHash: Core.ObjectHash, coreObjects: Core.DataObject,
+                rootHash: Core.ObjectHash, coreObjects: object,
```
brollb, line 2598: I prefer a better type than `object`

typings/webgme.d.ts
```diff
@@ -3143,103 +3249,7 @@ declare namespace Core {
...
-        getOwnValidAspectNames(): GmeCommon.Name[];
-        getAspectMeta(): GmeCommon.Metadata;
-
-        /** MixIns */
-        getMixinPaths(): GmeCommon.Path[];
-        canSetAsMixin(mixinPath: GmeCommon.Path): boolean;
-        isReadOnly(): boolean;
-
-    }
+    export type Node = object;
```
brollb, line 3252: Hmm... This isn't great since it will result in the following code no longer throwing a type error:
```
const notANode = {};
const name = core.getAttribute(notANode, 'name')
```

typings/webgme.d.ts
brollb: In the short term, you can even just use `type` to alias it for now until we can actual define a meaningful type.

brollb (changes requested):
It mostly looks good to me. The only major issue I have is with `export type Node = object;` (see the comment for more info).

typings/webgme.d.ts
pmeijer: It's used wrongly throughout, but I think the intention was that it should represent the raw data that maps to a node. In other words the json objects stored in the monodb.

pmeijer: *It was

typings/webgme.d.ts
pmeijer: This is a good point and I agree. Do you know if there is a way to define a type w/o specifying anything about it, but at the same time constrain anyone from passing anything that isn't of that type.
These `Core.Node`s are only being returned from a core instance.

PR #1793, opened 2024-03-19
Strip away library name from Canvas element in MetaEditor when library viewed

### LINKED ISSUE #1792: MetaEditor should strip away the namespace when viewing a meta model from a library
In this example `taxonomy` shouldn't be displayed - which makes the view look like it did when it was created and much more readable.

image

pmeijer:
Closes #1792


kecso (approved):
looking good

PR #1802, opened 2024-07-18
WIP: Add export library action to tree-brower

pmeijer:


kecso (commented):
Seems fine, but I would wait with the push until the engine update is finalized - not necessarily released, but pushed for sure

PR #1815, opened 2026-06-22
Add project info UI for kind and metadata

kecso:
## Summary
- Add **Project info** dialog for editing kind, description, owner, and name (with read-only audit fields)
- Add **Kind** column to the projects list with inline editing
- Add footer **Kind** button that opens the same dialog
- Wire **Project info …** into the breadcrumb project menu

## Test plan
- [ ] Open projects list — kind column visible when extra info is expanded; inline edit saves on Enter
- [ ] Create/open a project — footer kind button shows current kind (or placeholder); opens Project info dialog
- [ ] Breadcrumb project menu → **Project info …** opens dialog; save kind/description/owner/name
- [ ] Verify import package flow still works unchanged (no historyio/import changes in this PR)

Made with [Cursor](https://cursor.com)


pmeijer (approved):
Looks good

